PR #9, opened 2023-12-09
Numeric predictors draft, part 1

topepo:
Content for the first part of the initial predictor chapter. The feature extraction bits will eventually follow in another PR.


chapters/numeric-predictors.qmd
```diff
@@ -0,0 +1,397 @@
...
+To start, we'll consier two classes of transformations for individual predictors: those that resolve distributional skewness and those that convert each predictor to a common distribution (or scale). 
+
+After these, an example of a _group_ transformation is described. 
+
+### Resolving skewness
+
+The skew of a distribution can be quantified using the skewness statistic: 
+
+$$\begin{align}
+  skewness &= \frac{1}{(n-1)v^{3/2}} \sum_{1=1}^n (x_i-\overline{x})^3 \notag \\
```
topepo, line 64:
```suggestion
  skewness &= \frac{1}{(n-1)v^{3/2}} \sum_{i=1}^n (x_i-\overline{x})^3 \notag \\
```

chapters/numeric-predictors.qmd
```diff
@@ -0,0 +1,397 @@
...
+  \text{where}\quad  v &= \frac{1}{(n-1)}\sum_{1=1}^n (x_i-\overline{x})^2 \notag
```
topepo, line 65:
```suggestion
  \text{where}\quad  v &= \frac{1}{(n-1)}\sum_{i=1}^n (x_i-\overline{x})^2 \notag
```

chapters/numeric-predictors.qmd
topepo: suggestion from @krz

chapters/numeric-predictors.qmd
topepo: suggestion from @krz

chapters/numeric-predictors.qmd
```diff
@@ -0,0 +1,387 @@
...
+#| include: false
+source("../R/setup_ames.R")
+```
+
+As mentioned previously, feature engineering is the process of representing your predictor data so that the model has to do the least amount of work to explin the outcome effectively. There is also the need to properly encode/format the data based on the model's mathematical requirements (i.e., pre-processing). The next chapter described techniques for categorical data, and in this chapter, we do the same for quantitative predictors. 
+
+We'll begin with operations that only involve one predictor at a time before moving on to group transformations. Later, in @sec-add-remove-features,  procedures on numeric predictors are described that create additional predictor columns from a single column, such as basis function expansions. However, this chapter focuses on transformations that leave the predictors "in place" but altered. 
+
+
+## When are transformations estimated and applied?
```
kjell-stattenacity, line 41:
```suggestion
## What are Problematic Characteristics, and When Should Transformations be Applied?
```

chapters/numeric-predictors.qmd
```diff
@@ -0,0 +1,387 @@
...
+
+We'll begin with operations that only involve one predictor at a time before moving on to group transformations. Later, in @sec-add-remove-features,  procedures on numeric predictors are described that create additional predictor columns from a single column, such as basis function expansions. However, this chapter focuses on transformations that leave the predictors "in place" but altered. 
+
+
+## When are transformations estimated and applied? 
+
+The next few chapters concern preprocessing and feature engineering tools that mostly affect the predictors. As previously noted, the training set data are used to estimate parameters; this is also true for preprocessing parameters. All of these computations use the training set. At no point do we re-estimate parameters when new data are encountered.
```
kjell-stattenacity, line 43:
```suggestion
Common problematic characteristics that occur across individual predictors are:

* skewed or unusually shaped distributions,
* sample(s) that have extremely large or small values, and
* vastly disparate scales.

 Some models, like those that are tree-based, are able to tolerate these characteristics.  However, these characteristics can detrimentally affect most other models.  Techniques used to address these problems generally involve transformation parameters.  For example, to place the predictors on the same scale, we would subtract the mean of a predictor from a sample and then divide by the standard deviation.  This is know as standardizing and will be discussed in the next section.   
```

chapters/numeric-predictors.qmd
```diff
@@ -0,0 +1,387 @@
...
+
+
+## When are transformations estimated and applied? 
+
+The next few chapters concern preprocessing and feature engineering tools that mostly affect the predictors. As previously noted, the training set data are used to estimate parameters; this is also true for preprocessing parameters. All of these computations use the training set. At no point do we re-estimate parameters when new data are encountered. 
+
+For example, a standardization tool that centers and scales the data is introduced in the next section. The mean and standard deviation are computed from the training set for each column being standardized. When the training set, test set, or any future data are standardized, it uses these statistics derived from the training set. Any model fit that uses these standardized predictors would want new samples being predicted to have the same reference distribution.
```
kjell-stattenacity, line 45:
```suggestion
What data should be used to estimate the mean and standard deviation?  Recall, the training data set was used to estimate model parameters.  Similarly, we will use the training data to estimate transformation parameters.  When the test set or any future data set are standardized, the process will use the estimates from the training data set.  Any model fit that uses these standardized predictors would want new samples being predicted to have the same reference distribution.
```

chapters/numeric-predictors.qmd
```diff
@@ -0,0 +1,387 @@
...
+
+For example, a standardization tool that centers and scales the data is introduced in the next section. The mean and standard deviation are computed from the training set for each column being standardized. When the training set, test set, or any future data are standardized, it uses these statistics derived from the training set. Any model fit that uses these standardized predictors would want new samples being predicted to have the same reference distribution. 
+
+Suppose that a predictor column had an underlying Gaussian distribution with a sample mean estimate of 5.0 and a sample standard deviation of 1.0. Suppose a new sample has a predictor value of 3.7. For the training set, this new value lands around the 10th percentile and would be standardized to a value of -1.3. The new value is relative to the training set distribution. Also note that, in this scenario,  it would be impossible to standardize using a recomputed standard deviation for the new sample (which means we try to divide with a zero standard deviation).
```
kjell-stattenacity, line 47:
```suggestion
Suppose that a predictor column had an underlying Gaussian distribution with a sample mean estimate of 5.0 and a sample standard deviation of 1.0. Suppose a new sample has a predictor value of 3.7. For the training set, this new value lands around the 10th percentile and would be standardized to a value of -1.3. The new value is relative to the training set distribution. Also note that, in this scenario,  it would be impossible to standardize using a recomputed standard deviation for the new sample (which means we try to divide with a zero standard deviation). 

Now let's review transformations that leave the predictor "in-place", but altered.
```

chapters/numeric-predictors.qmd
```diff
@@ -0,0 +1,387 @@
...
+
+Suppose that a predictor column had an underlying Gaussian distribution with a sample mean estimate of 5.0 and a sample standard deviation of 1.0. Suppose a new sample has a predictor value of 3.7. For the training set, this new value lands around the 10th percentile and would be standardized to a value of -1.3. The new value is relative to the training set distribution. Also note that, in this scenario,  it would be impossible to standardize using a recomputed standard deviation for the new sample (which means we try to divide with a zero standard deviation). 
+
+## General Transformations
+
+Many transformations that involve a single predictor change the data distribution. What would a problematic distribution look like? Most predictive models do not place specific parametric assumptions on the predictor variables (e.g., require normality), but some distributions might facilitate better predictive performance than others.
```
kjell-stattenacity, line 51:
```suggestion
Many transformations that involve a single predictor change the data distribution. Most predictive models do not place specific parametric assumptions on the predictor variables (e.g., require normality), but some distributions might facilitate better predictive performance than others. 
```

kjell-stattenacity (changes requested):
Suggested edits

PR #11, opened 2023-12-09
Categorical predictors chapter draft

topepo:
A fairly comprehensive draft of this chapter (which would come after the numeric predictor chapter)


R/setup_hotel_rates.R
```diff
@@ -1,28 +1,7 @@
...
-hotel_rate_test_range <- format(hotel_rate_test_range, "%e %B %Y")
-
-hotel_rate_test <-
-  hotel_rate_test %>%
-  select(-arrival_date)
+hotel_rate_test  <- testing(hotel_rate_split)
 
 set.seed(6858)
-hotel_rate_rs  <- validation_split(hotel_rate_train, strata = avg_price_per_room)
+hotel_rate_rs  <- vfold_cv(hotel_rate_train, strata = avg_price_per_room)
```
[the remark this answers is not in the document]
topepo: Nevermind. I figured out a good time series resampling scheme and updated the file.

chapters/categorical-predictors.qmd
```diff
@@ -0,0 +1,718 @@
...
+
+Why would we need to modify this type of predictor? First, most mathematical models require information to be numbers. We'll need a method to convert qualitative data effectively to some quantitative format. This is an encoding problem; we are finding another representation of our predictor data. There are various tools for converting qualitative data, ranging from simple indicator variables (a.k.a. dummy variables) to more statistically sophisticated tools. These are the focus of this chapter. 
+
+Second, there are other circumstances where simplifying or transforming categorical predictors might be advantageous. These are usually related to exceptional situations, such as when there are many possible values (i.e., high cardinality) or infrequently occurring values. Techniques to address these issues are also discussed. 
+
+As implied above, some modeling techniques can take the categorical predictors in their natural form. These include tree- or rule-based models, naive Bayes models, and others. In this book's third and fourth parts, we'll discuss these details in their respective sections.
```
kjell-stattenacity, line 101:
```suggestion
As implied above, some modeling techniques can take the categorical predictors in their natural form. These include tree- or rule-based models, naive Bayes models, and others. In this book's third and fourth parts, we'll discuss how certain models naturally handle categorical variables in their respective sections.
```

chapters/categorical-predictors.qmd
```diff
@@ -0,0 +1,718 @@
...
+
+## Example Data: Daily Hotel Rates {#sec-hotel-rates}
+
+@ANTONIO201941 describe a data set with booking records from two Portuguese hotels. There are numerous variables in the data related to the
+
+ - Reservation: room type, date, meal plan, etc., 
+ - Booking method: agent, distribution channel, etc., 
+ - Customer: type, number of children staying, country of origin, etc. 
+
+There are several qualitative variables: customer type, market segment, room type, etc.
```
kjell-stattenacity, line 114:
```suggestion
In addition, there are several qualitative variables: customer type, market segment, room type, etc. 
```

chapters/categorical-predictors.qmd
```diff
@@ -0,0 +1,718 @@
...
+  bake(new_data = NULL) %>% 
+  arrange(customer_type) %>% 
+  mutate(customer_type = gsub("_", " ", customer_type)) %>% 
+  rename_all(~ gsub("_", " ", .x)) %>% 
+  gt() %>% 
+  tab_spanner(label = "Indicator Columns", columns = c(-`customer type`)) %>% 
+  tab_options(table.width = pct(50))
+```
+
+The rationale for excluding one column is that you can infer the reference value if you know the values of all of the existing indicator columns^[In other words, we know that the vector of indicators (0, 0, 0) must represent the contract customers.]. Including all possible indicator columns embeds a redundancy in the data.
```
kjell-stattenacity, line 158:
```suggestion
The rationale for excluding one column is that you can infer the reference value if you know the values of all of the existing indicator columns^[In other words, we know that the vector of indicators (0, 0, 0) must represent the contract customers.]. Including all possible indicator columns embeds a redundancy in the data.  As we will see shortly, data that contain this type of redundancy pose problems for some models like linear regression.
```

chapters/categorical-predictors.qmd
```diff
@@ -0,0 +1,718 @@
...
+
+An example of this is the one-hot encoding method shown in @tbl-one-hot. If we were to include an intercept term with a one-hot encoding, the row-wise sum of the indicator columns equals the intercept column. The implication is that one of the model terms is not "estimable." For this reason, this encoding method is mainly used with models that do not use specific linear algebra operations (e.g., matrix inversion) to estimate parameters. Some tools can identify linear combinations [@estimable], and from this information, the smallest possible set of columns can be found to delete and thereby eliminate the linear dependencies. 
+
+Note that linear dependencies are exact linear relationships between predictors. These are other circumstances where two or more predictors are highly correlated but not perfectly related. This situation, called **multicolinearity**, is discussed throughout the book but mostly in @sec-colinearity.
```
kjell-stattenacity, line 211:
```suggestion
Note that linear dependencies are exact linear relationships between predictors. There are other circumstances where two or more predictors are highly correlated but not perfectly related. This situation, called **multicolinearity**, is discussed throughout the book but mostly in @sec-colinearity.  Multicolinearity can be a problem for many models and needs to be resolved prior to modeling. 
```

chapters/categorical-predictors.qmd
```diff
@@ -0,0 +1,718 @@
...
+
+In addition to reference cell and one-hot encodings, there are various tools for creating indicator columns. Some, called _contrast methods_, are associated with more niche strategies, and many produce fractional values instead of binary indicators. For example, Helmert contrasts [@ruberg] are configured so that each category is compared to the average of the preceding categories. No encoding method is uniformly better than the others; the choice is often driven by how the user will interpret the parameter estimates. 
+
+The situation becomes slightly more complex when a model contains multiple categorical predictors. The reference value combines levels across predictors for reference value parameterizations. For example, if a model included the customer's country of residence, the reference would include that factor. For example, if Canadians were the reference nationality, the overall reference cell would be Canadian contract customers.
```
kjell-stattenacity, line 215:
```suggestion
The situation becomes slightly more complex when a model contains multiple categorical predictors. The reference value combines levels across predictors for reference value parameterizations. For example, if a model included the customer's country of residence, the reference would include that factor. Suppose Canadians were the reference nationality.  Then the overall reference cell would be Canadian contract customers. 
```

chapters/categorical-predictors.qmd
```diff
@@ -0,0 +1,718 @@
...
+
+The situation becomes slightly more complex when a model contains multiple categorical predictors. The reference value combines levels across predictors for reference value parameterizations. For example, if a model included the customer's country of residence, the reference would include that factor. For example, if Canadians were the reference nationality, the overall reference cell would be Canadian contract customers. 
+
+Note that these encoding techniques are basically look-up tables that map individual categories known in the training set to specific numeric values. As discussed in the next section, they can be problematic when novel values are seen in new data (such as the test set).
```
kjell-stattenacity, line 217:
```suggestion
Note that these encoding techniques are basically look-up tables that map individual categories known in the training set to specific numeric values. As discussed in the next section, mapping categories can be problematic when novel values are seen in new data (such as the test set). 
```

chapters/categorical-predictors.qmd
```diff
@@ -0,0 +1,718 @@
...
+
+Note that these encoding techniques are basically look-up tables that map individual categories known in the training set to specific numeric values. As discussed in the next section, they can be problematic when novel values are seen in new data (such as the test set). 
+
+## Novel Categories {#sec-novel-categories}
+
+When creating indicator variables, there is the assumption that the categories in the training data reflect all possible categories. This may not be the case. For example, the agents observed in our data are not the only agents that will be booking rooms in the future. After we create a model with a specific set of agents, what happens when we predict future data that involves new agents? 
+
+One approach we could take is to prospectively create a new value for the predictor (let's give it a value of `"new"`). When we have new agents, we could assign them to this general category so that the model can anticipate this new level. In general, this won't help since the model estimation process will never have any "new" values on hand when the model is fit. For some models, this is innocuous and has no real harm (but offers no benefit either).
```
kjell-stattenacity, line 223:
```suggestion
One approach we could take is to prospectively create a new value for the predictor (let's give it a value of `"new"`). When a future sample has an agent that was not in the original data, we could assign this sample to the general category so that the model can anticipate this new level. In general, this won't help since the model estimation process will never have any "new" values on hand when the model is fit. For some models, this is innocuous and has no real harm (but offers no benefit either). 
```

chapters/categorical-predictors.qmd
```diff
@@ -0,0 +1,718 @@
...
+freq_cut <- 1 / 20
+low_nums <- 
+  hotel_rate_train %>% 
+  count(agent) %>% 
+  mutate(pct = n / nrow(hotel_rate_train) * 100) %>% 
+  filter(pct < freq_cut) %>% 
+  nrow()
+```
+
+It is possible to isolate infrequently occurring categories into a new "other" category. For example, there are `r low_nums` agents with fewer bookings than `r freq_cut`{{< pct >}} of the training set. It may make sense to group these agents into their own categories as long as the predictive ability of the variable is maintained. There is no magic number for the training set frequency cutoff that decides which categories to pool. This pre-processing value requires optimization.
```
kjell-stattenacity, line 269:
```suggestion
It is possible to isolate infrequently occurring categories into a new "other" category. For example, there are `r low_nums` agents with fewer bookings than `r freq_cut`{{< pct >}} of the training set. It may make sense to group these agents into their own categories as long as the predictive ability of the variable is maintained. There is no magic number for the training set frequency cutoff that decides which categories to pool. This pre-processing value requires optimization since each data set will have its own nuances with infrequent categories.
```

chapters/categorical-predictors.qmd
```diff
@@ -0,0 +1,718 @@
...
+            locations = cells_body(rows = 10)) %>% 
+  cols_align(
+    align = "right",
+    columns = everything()
+  ) %>% 
+  tab_options(table.width = pct(70))
+
+```
+
+The main downside of this method is that the use of hash values makes it impossible to explain the model. If the tenth feature column is critical, we can't explain why this is the case for new data (since the hash function is practically non-reversible and may include collisions).  This may be fine if the primary objective is prediction rather than interpretation.  We also need to tune the number of hashing columns to optimize performance.
```
kjell-stattenacity, line 375:
```suggestion
The main downside of this method is that the use of hash values makes it impossible to explain the model. If the tenth feature column is critical, we can't explain why this is the case for new data (since the hash function is practically non-reversible and may include collisions).  This may be fine if the primary objective is prediction rather than interpretation.  When the goal is to optimize predictive performance, then the number of hashing columns to use can be included as a tuning parameter.  The model tuning process can then determine an optimal value of the number of hashing columns.
```

chapters/categorical-predictors.qmd
```diff
@@ -0,0 +1,718 @@
...
+
+How well can we estimate an effect if there are only one or two bookings for an agent in the data set?
+
+Both of these concerns can be addressed by _how_ we estimate the effect of the predictor. Recall that @tbl-one-hot showed how simple linear regression was used to calculate the mean for each customer type. The way that the indicator variables impact the parameter estimates is such that the rows for each customer type are the only data used to estimate that customer type's effect. For example, the contract data do not directly affect the parameter estimate for the transient customers, and so on. This is often called a _no pooling_ method of estimating parameters. In this context, "pooling" means combining data across multiple categories to estimate parameters.  
+
+We could compute the naive ADR estimate using sample means and produce a similar table. We can create better mean estimates for each agent by considering the data from _all_ agents using a technique called _partial pooling_ [@gelman1995bayesian; @BayesRules,  chap. 15].
```
kjell-stattenacity, line 453:
```suggestion
We could compute the naive ADR estimate using sample means and produce a similar table. However, we can create better mean estimates for each agent by considering the data from _all_ agents using a technique called _partial pooling_ [@gelman1995bayesian; @BayesRules,  chap. 15]. 
```

chapters/categorical-predictors.qmd
```diff
@@ -0,0 +1,718 @@
...
+#### Numeric outcomes {-}
+
+For our ADR analysis, the outcome is a quantitative value. A linear regression model (see @sec-linear-regression) is the most commonly used approach to model this type of outcome. Linear regression models the outcome as a function of a slope and intercept model. For our example, where we only consider estimating the effect of the agents using indicator columns, the model is:
+
+$$
+y_{ij} =\beta_0 + \beta_1x_{i1} + \ldots + \beta_{`r nrow(blups) -1`}x_{i `r nrow(blups) -1`} + \epsilon_{ij} = \mu_i  + \epsilon_{ij}
+$$
+
+where $y_{ij}$ is the outcome (in this case, the ADR of booking $i$ by agent $j$),  $x_{ij}$ is an indicator for reservation $i$ from agent $j$,  $\beta_0$ is the estimated outcome of the reference cell, and we assume $\epsilon_i \sim N(0, \sigma^2_\epsilon)$. This leads to the assumption that $y_{ij} \sim N(\mu_i, \sigma^2_\epsilon)$. 
+ The $\hat{\beta}_{1_j}$ parameters are the effect of the agents above and beyond the reference cell effect. The index $j$ is from 1 to `r nrow(blups) -1` since there were `r nrow(blups)` agents in the training set with at least one booking.
```
kjell-stattenacity, line 468: Should $\hat{\beta}_{1_j}$ be $\hat{\beta}_{j}$. ?

chapters/categorical-predictors.qmd
```diff
@@ -0,0 +1,718 @@
...
+  pivot_longer(cols = c(starts_with("Feature")), names_to = "variable", values_to = "value") %>% 
+  ggplot(aes(x = quality, y = value)) + 
+  geom_point(aes(col = quality), cex = 2.5, show.legend = FALSE) + 
+  geom_line(aes(x = ordering), alpha = 1 / 3) + 
+  facet_wrap(~ variable) +
+  scale_color_brewer(palette = "Set2") +
+  labs(y = NULL)
+```
+
+In many applications, higher-degree polynomials (> 2) are unlikely to be effective predictors. For example, the cubic pattern in @fig-ordered ("Feature 3") implies that houses in Ames with '`excellent`' quality heating systems would have lower sale prices than those with '`fair`' or '`poor`' systems.
```
kjell-stattenacity, line 716:
```suggestion
In many applications, higher-degree polynomials (> 2) are unlikely to be effective predictors. For example, the cubic pattern in @fig-ordered ("Feature 3") implies that houses in Ames with '`excellent`' quality heating systems would have lower sale prices than those with '`fair`' or '`poor`' systems.  Utilizing higher order polynomials usually requires some specific knowledge about the problem at hand.
```

kjell-stattenacity (changes requested):

chapters/categorical-predictors.qmd
topepo: Yep! Good eye!

Issue #4, opened 2023-11-22
Better styling for callout boxes


topepo:
I don't like the quarto default: 

<img width="606" alt="image" src="https://github.com/aml4td/website/assets/5731043/7b8fee04-370d-4699-b2b9-dbe66538fe28">

The ones in _Interactively exploring high-dimensional data and models in R_ ([examples](https://dicook.github.io/mulgar_book/4-pca.html)) are beautiful and Mine's [class notes](https://openintro-ims2.netlify.app/17-inference-two-props) ones are also an improvement.

krz:
What do you have in mind?
Something with an image like in your first link? Which colors?

here is something you can try in [https://jsfiddle.net/](https://jsfiddle.net/)
![image](https://github.com/aml4td/website/assets/210623/370ca4c6-ab6a-4468-b127-a3bf404f9659)

html:
```
<div class="insight">
Some content.
</div>
```

css:
```
.insight {
  padding: 2em;
  padding-left: 100px;
  margin: 2em 0;
  background-color: #e2ecb0;
  background-image: url("https://i.ibb.co/zVFnfq1/leaf-clean.png");
  background-size: 60px;
  background-position: 10px 1em;
  background-repeat: no-repeat;
  border-right: 6px solid #5b6d3c;
  position: relative;
  font-size: 0.9em;
  border-radius: 3px;
}
```

topepo:
Looking around, something like [these](https://dribbble.com/shots/12883156-Callout-component-Ui-Figma-React-design-system) would be good. I like the solid color in the header and body along with the bevel.

krz:
maybe something like this:
![image](https://github.com/aml4td/website/assets/210623/5605b5b6-646c-42f8-9529-56214ef1109a)

html:
```
<div class="importantinfo">
Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut labore et dolore magna aliquyam erat, sed diam voluptua.
</div>
```

css:
```
@import url('https://fonts.googleapis.com/css2?family=Libre+Franklin:wght@100;300&display=swap');

.importantinfo {
  padding: 0.25em 1em 1em 3em;
  margin-top: 20px;
  margin-bottom: 20px;
  background: #a9d3ab 5px center/3em no-repeat;
  background-image: url("https://svgshare.com/i/11W1.svg");
  background-position: 0.6em 0.5em;
  background-size: 3% auto;
  border-radius: 10px;
[...]

PR #28, opened 2024-01-05
add callout box "important information"

krz:
discussed in #4, this PR adds css for a callout box "important information" that can be accessed through:
```
::: {.importantinfo}
Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut labore et dolore magna aliquyam erat, sed diam voluptua.
:::
```

image


topepo:
Thanks! I applied it to the existing cases. We can adapt this for the "warning" and "note" instances too.

PR #29, opened 2024-01-18
multidimensional scaling

topepo:
Significant sections on multidimensional scaling, framing feature extraction as embeddings. The focus is on isomap and umap. There’s also a shiny-live app at the end. 

I skipped over linear feature extraction and autoencoders (assuming we want to talk about them in depth)

#9 should be merged prior to this PR.


chapters/embeddings.qmd
```diff
@@ -0,0 +1,646 @@
...
+
+There are many different MDS methods. @Ghojogh2023 has an excellent review of them and their nuances. 
+
+Some MDS methods compute all pairwise distances between points and use this as the input to the embedding algorithm. This is similar to how PCA can be estimated using the covariance or correlation matrices. One technique, _Non-Metric MDS_ [@kruskal1964multidimensional;@kruskal1964nonmetric;@sammon1969nonlinear], finds embeddings that minimize an objective function called "stress":
+
+$$
+\text{Stress} = \sqrt{\frac{\sum\limits^{n_{tr}}_{i = 1}\;\sum\limits^{n_{tr}}_{j = i+1}\left(d(x_i, x_j) - d(x^*_i, x^*_j)\right)^2}{\sum\limits^{n_{tr}}_{i = 1}\;\sum\limits^{n_{tr}}_{j = i+1}d(x_i, x_j)^2}}
+$$
+
+The denominator uses the squared difference between the pairwise distances in the original values ($x$) and the smaller embedded dimension ($x^*$). The summations only move along one diagonal of the distance matrices to reduce redundant computations. @fig-mds-example(c, top row) has the resulting one dimensional projection of our two-dimensional data.
```
kjell-stattenacity, line 315: I think denominator should be numerator?  Also, I'm not following the sentence "The summations only move along one diagonal..."  The formula as written sums across all pairs of points.

chapters/embeddings.qmd
```diff
@@ -0,0 +1,646 @@
...
+
+This can be an effective dimension reduction procedure, although there are a few issues. First, the entire matrix of distances is required (with $n_{tr}(n_{tr}-1)/2$ entries). For large training sets, this can be unwieldy and time-consuming.  Second, like PCA, it is a global method. We might be able to achieve more nuanced embeddings by focusing on local structures. Finally, metric MDS is not easily made to be projective; how can we apply the same process to new data?
```
kjell-stattenacity, line 317:
```suggestion
This can be an effective dimension reduction procedure, although there are a few issues. First, the entire matrix of distances is required (with $n_{tr}(n_{tr}-1)/2$ entries). For large training sets, this can be unwieldy and time-consuming.  Second, like PCA, it is a global method that uses all data in the computations.  We might be able to achieve more nuanced embeddings by focusing on local structures.  Finally, it is challenging to apply metric MDS to project new data onto the space in which the original data was projected.
```

kjell-stattenacity (changes requested):

PR #32, opened 2024-02-12
Interactions

topepo:
A fairly full section on interactions and their detection. 

Notes to follow. 

We use a shinylive app 💅


chapters/interactions-nonlinear.qmd
```diff
@@ -0,0 +1,893 @@
...
+
+The healthcare industry has long understood the concept of interactions among drugs for treating specific diseases [@singh2017suppressive, @mokhtari2017combination, and @altorki2021neoadjuvant].  As an example of an interaction, consider treatment for the disease non-small cell lung cancer (NSCLC).  In a recent study, patients with an advanced stage of NSCLC with an EGFR mutation were given either osimertinib alone or osimertinib in combination with traditional chemotherapy [@tragrisso2023].   Patients taking the combination treatment had a significantly longer progression-free survival time than patients taking osmertinib alone.  Hence, the interaction of the treatments is more effective than the single treatment.  To summarize, two (or more) predictors interact if their combined impact is different (less or greater) than what would be expected from the added impact of each predictor alone.
```
topepo, line 162: @kjell-stattenacity I don't have the bib entry for `tragrisso2023`.

chapters/interactions-nonlinear.qmd
```diff
@@ -0,0 +1,893 @@
...
+ - A third scenario is when $\beta_3$ is essentially zero.  In this case, there is no interaction between the predictors and the relationship between the predictors is _additive_. 
+ - Finally, for some data sets, we may find that neither predictor is important individually (i.e., $\beta_1$ and $\beta_2$ are zero).  However, the coefficient on the interaction term is not zero.  Because this case occurs very infrequently, it is called _atypical_. 
+ 
+To understand this better, @fig-interaction-contours shows a contour plot of a predicted linear regression model with various combinations of the model slope parameters. The two predictors are centered at zero with values ranging within $x_j \pm 4.0$). The default setting shows a moderate synergistic interaction effect since all of the $\beta_j = 1.0$). In the plot, darker values indicate smaller predicted values.   
+
+::: {#fig-interaction-contours}
+
+::: {.figure-content}
+
+```{shinylive-r}
```
topepo, line 292: I absolutely love this.

chapters/interactions-nonlinear.qmd
```diff
@@ -0,0 +1,893 @@
...
+#| label: fig-interaction-contours
+#| viewerHeight: 600
+#| standalone: true
+
+library(shiny)
+library(ggplot2)
+library(bslib)
```
topepo, line 299: Our shiny folks helped me with a lot of formatting here. It's new syntax but it looks a lot better and will reshape itself well with very small screens.

chapters/interactions-nonlinear.qmd
```diff
@@ -0,0 +1,893 @@
...
+
+ui <- page_fillable(
+  theme = bs_theme(bg = "#fcfefe", fg = "#595959"),
+  padding = "1rem",
+  layout_columns(
+    fill = FALSE,
+    col_widths = breakpoints(xs = c(-2, 8, -2), sm = 4),
```
topepo, line 306: Sooo... the standard is to assume that there are 12 units of space to play with. `xs` means extra small and `sm` is small.  We have one plot element and `c(-2, 8, -2)` means that allocate 8 columns (of 12) to it (with 2 empty columns on either side).

chapters/interactions-nonlinear.qmd
```diff
@@ -0,0 +1,893 @@
...
+  ggplot(aes(time_to_delivery, .pred)) +
+  geom_abline(col = "green", lty = 2) +
+  geom_point(alpha = 0.4, cex = 1) +
+  geom_smooth(se = FALSE, col = "red") +
+  coord_obs_pred() +
+  facet_wrap( ~ Model, ncol = 3) +
+  labs(x = "Observed Delivery Time", y = "Predicted Delivery Time")
+```
+
+So far, we have used visualizations to unearth potential interactions. This can be an effective strategy when the number of potential interactions is large, but the visual nature of this process is subjective. There are some specialized quantitative tools for identifying interactions. For example, @lim2015learning used regularized generalized linear models (sections TODO) to estimate all possible two-way interactions and use a penalization method to determine which should be retained. @miller1984selection and @fes ([Section 7.4.3](https://bookdown.org/max/FES/approaches-when-complete-enumeration-is-practically-impossible.html#the-feasible-solution-algorithm)) describe the feasible solution algorithm, an iterative search method for interaction discovery. Another, which we will now describe in more detail, is Friedman's H-statistic [@friedman2008predictive].
```
topepo, line 620: @kjell-stattenacity is `miller1984selection` the best reference for FSA and interactions?

chapters/interactions-nonlinear.qmd
```diff
@@ -0,0 +1,893 @@
...
+  scale_color_manual(values = day_cols)
+
+(((p_hour_ind + p_hour) / (p_day + p_hour_day)) +
+  plot_layout(guides = "collect")) &
+  theme(legend.position = "bottom")
+```
+
+Friedman’s H-statistic can quantify the effect of interaction terms using partial dependence profiles. When investigating two-factor interactions between predictors $j$ and $j'$, the statistic is 
+
+$$
```
topepo, line 758: There is a lot of abuse of notation here since $f(x_{ij})$ is really a vector. Hopefully, the PDP in the figure helps clarify that.

chapters/interactions-nonlinear.qmd
```diff
@@ -0,0 +1,893 @@
...
+
+If meaningful interactions are unknown before modeling, can models still discover and utilize these potentially important features? Recent studies using various advanced modeling techniques have shown that some methods can inherently detect interactions. For example, tree-based models [@elith2008working], random forests [@garcia2009evaluating], boosted trees [@lampa2014identification], and support vector machines [@chen2008support] are effective at uncovering them. 
+
+If modern modeling techniques can naturally find and utilize interactions, then why is it necessary to spend any time uncovering these relationships?  The first reason is interpretability.  Recall the trade-off between prediction and interpretation that was discussed in Section TODO.  More complex models are less interpretable and generally more predictive, while simpler models are more interpretable and less predictive. A second reason is that, by identifying essential interactions, we can include them in simpler models to improve predictive performance and maintain easy interpretability.
```
kjell-stattenacity, line 421:
```suggestion
If modern modeling techniques can naturally find and utilize interactions, then why is it necessary to spend any time uncovering these relationships?  The first reason is to improve interpretability.  Recall the trade-off between prediction and interpretation that was discussed in Section TODO.  More complex models are less interpretable and generally more predictive, while simpler models are more interpretable and less predictive.   Therefore, if we know which interaction(s) are important, we can include these in a simpler model to enable a better interpretation.  A second reason to spend time uncovering interactions is to help improve the predictive performance of models.  
```

chapters/interactions-nonlinear.qmd
```diff
@@ -0,0 +1,893 @@
...
+
+added_all_pairs <- anova_res$`# Added`[3]
+impr_add_two <- sprintf("%3.2f", (anova_res$Decimal[1] - anova_res$Decimal[2]))
+impr_add_all <- sprintf("%3.2f", (anova_res$Decimal[2] - anova_res$Decimal[3]))
+```
+
+ 
+What should we do if we already have candidate interaction sets?  If we use a more formal statistical model, such as linear or logistic regression, the traditional tool for evaluating whether additional model terms in a set of nested models are worth including is the conventional analysis of variance (ANOVA). This uses the statistical likelihood value as the objective function. This is equivalent to comparing the RMSE values between models for linear regression ^[It so happens that the Gaussian likelihood function is equivalent to the sums of squared errors (SSE). That, in turn, is equivalent to the RMSE. That simplification does not automatically occur for other probability distributions.]. The error reduction and how many additional terms are responsible for the improvement are computed. Using these results and some probability assumptions about the data, we can formally test the null hypothesis that the additional parameters all have coefficient values of zero.
```
kjell-stattenacity, line 538:
```suggestion
What should we do if we have a candidate set of interactions?  If we are using a more formal statistical model, such as linear or logistic regression, the traditional tool for evaluating whether additional model terms in a set of nested models are worth including is the conventional analysis of variance (ANOVA). This uses the statistical likelihood value as the objective function, which is equivalent to comparing the RMSE values between models for linear regression ^[It so happens that the Gaussian likelihood function is equivalent to the sums of squared errors (SSE). That, in turn, is equivalent to the RMSE. That simplification does not automatically occur for other probability distributions.]. The error reduction and how many additional terms are responsible for the improvement are computed. Using these results and some probability assumptions about the data, we can formally test the null hypothesis that the additional parameters all have coefficient values of zero. 
```

chapters/interactions-nonlinear.qmd
```diff
@@ -0,0 +1,893 @@
...
+
+Looking at the delivery data, our model in @sec-model-development-whole-game included a single set of interactions (e.g., hour-by-day). What if we included one more interaction: item 1 $\times$ item 10? @tbl-interaction-anova shows the ANOVA results. The RMSE _computed on the training set_ is listed in the first and second rows. The reduction by including this additional model term is `r impr_add_two` (decimal minutes). Assuming normality of the model residuals, the p-value for this test is exceedingly small. This indicates that there is no evidence that this parameters is truly zero. This statistical difference may not make much of a practical difference. However, machine learning models often behave like "a game of inches" where every small improvement adds up to an overall improvement that matters.
```
kjell-stattenacity, line 540:
```suggestion
Looking at the delivery data, our model in @sec-model-development-whole-game included a single set of interactions (e.g., hour-by-day). What if we included one more interaction: item 1 $\times$ item 10? @tbl-interaction-anova shows the ANOVA results. The RMSE _computed on the training set_ is listed in the first and second columns. The reduction by including this additional model term is `r impr_add_two` (decimal minutes). Assuming normality of the model residuals, the p-value for this test is exceedingly small. This indicates that there is no evidence that this parameter is truly zero. In other words, there is strong evidence that the inclusion of the interaction helps explain the response.  This statistical difference may not make much of a practical difference. However, machine learning models often behave like "a game of inches" where every small improvement adds up to an overall improvement that matters. 
```

chapters/interactions-nonlinear.qmd
```diff
@@ -0,0 +1,893 @@
...
+
+One problem with the ANOVA method is that it calculates the model error using the training set, which we know may not indicate what would occur with previously unseen data. In fact, it is well known that, for linear regression via ordinary least squares estimation, it is impossible for the training set error to ever increase when adding new model terms. You can see this in @tbl-interaction-anova where the validation set RMSE and MAE values are included. Note that, when the large set of interactions were added, these metrics both _increase_ in the validation set, a result contrary to what the ANOVA results tell us.
```
kjell-stattenacity, line 579:
```suggestion
One problem with the ANOVA method is that it calculates the model error using the training set, which we know may not indicate what would occur with previously unseen data. In fact, it is well known that, for linear regression via ordinary least squares estimation, it is impossible for the training set error to ever increase when adding new model terms. Therefore, was the drop in RMSE when all interactions were included due to this fact?  Or was it due to other important interactions that were included?  To understand this we can turn to the validation set to provide confirmation.  You can see this in @tbl-interaction-anova where the validation set RMSE and MAE values are included. Note that, when the large set of interactions were added, these metrics both _increase_ in the validation set, a result contrary to what the ANOVA results tell us.  
```

chapters/interactions-nonlinear.qmd
kjell-stattenacity: Where did we originally write this example?  I just can't put my finger on it.

chapters/interactions-nonlinear.qmd
kjell-stattenacity: I need to check on this

kjell-stattenacity (changes requested):

chapters/interactions-nonlinear.qmd
topepo: Added a new issue for this to be able to merge this PR

PR #37, opened 2024-04-02
Splines and polynomial basis expansions

topepo:
New sections. Fairly heavy use of shinylive


topepo:
For reviewers who don't/can't build the site, here is a crude export of the new materials [(new_section_splines.pdf)](https://github.com/aml4td/website/files/14841317/new_section_splines.pdf). No interactivity (obviously) but the text is fully available.

kjell-stattenacity (changes requested):

PR #42, opened 2024-05-23
variance/bias tradeoff and Overfitting

topepo:


chapters/interactions-nonlinear.qmd
```diff
@@ -789,6 +789,195 @@ The H-statistic and its alternatives can be an incredibly valuable tool for lear
...
+```
+
+
+```{r}
+#| label: variance-bias-stats
+#| echo: false
+#| cache: true
+set.seed(614)
+sim_res <- 
+  map_dfr(rep(1:23, each = 50000), sim) %>%
```
topepo, line 884: This will take a while to run the first time

chapters/overfitting.qmd
```diff
@@ -0,0 +1,487 @@
...
+load("../RData/bayes_opt_calcs.RData")
+```
+
+This chapter describes the most crucial concept in predictive modeling: overfitting. When a model fails, it is almost always due to overfitting of some sort. The problem is that you may only realize that there is an issue once you finish the development phase of modeling and are exposed to completely new data.
+
+To get started, we'll discuss model complexity and how it can be both good and bad. 
+
+## Model Complexity and Overfitting {#sec-complexity}
+
+Many predictive models have parameters. For example, the linear regression model initially shown in @eq-poly-linear-reg contains slopes and intercepts (the $\beta$ values) estimated during model training using the predictor and outcome data. In other words, it is possible to write the equations to solve to estimate these values.
```
topepo, line 49: @eq-poly-linear-reg  is in PR #37

chapters/interactions-nonlinear.qmd
```diff
@@ -789,6 +789,195 @@ The H-statistic and its alternatives can be an incredibly valuable tool for lear
 
 ## Discretization
 
+
+## Sidebar: The Variance-Bias Tradeoff {#sec-variance-bias}
+
+Our discussion of basis expansions presents an excellent time for a segue to discuss an essential idea in statistics and modeling: the variance-bias tradeoff. This idea will be relevant in upcoming sections for resampling and specific models. At the end of this section, we’ll also connect it to @sec-effect-encodings. 
+What does variance mean in terms of a machine learning model? In this context, it would quantify how much the fitted model changes if we slightly change the data. For example, would the curve change much if we repeat the data collection that produces the values in @sec-polynomials and fit the sample model (for a fixed sample size and polynomial degree)? We can also look at the _variance of prediction_: for a specific new data point $x_0$, how much intrinsic uncertainty is there? Similar to the discussion regarding extrapolation of global polynomials, we might want to compare how much the uncertainty changes as we move outside the training set’s range.
```
topepo, line 796: Some references here are to sections/equations in PR https://github.com/aml4td/website/pull/37

chapters/overfitting.qmd
```diff
@@ -0,0 +1,487 @@
...
+
+Similarly, the neural network required the number of hidden units to be defined (among other values). We weren't sure what that should be, so a set of candidate values were assessed, and the one with the best results was used for the final model fit (see @fig-neural-network). We've also seen specific tuning parameters in @sec-feature-hashing, @sec-combining-categories, and also for the embedding methods in @sec-embeddings. 
+
+In each of these cases, there was a **tuning parameter**  (a.k.a., _hyperparameter_) that defined the structure of the model and could not be directly estimated from the data. In each case, increasing the parameter's value made the model's potential structure more elaborate. The effect of this increase in the model's _complexity_ is additional capability to adapt to any pattern seen in the training data. The straight line fit associated with a linear regression model without polynomial terms would have done an abysmal job on such a nonlinear data set. Adding additional complexity allowed it to better conform to the training set data.
```
kjell-stattenacity, line 57:
```suggestion
In each of these cases, there was a **tuning parameter**  (a.k.a., _hyperparameter_) that defined the structure of the model and could not be directly estimated from the data. Increasing the parameter's value for each model made the model's potential structure more elaborate. The effect of this increase in the model's _complexity_ is additional capability to adapt to any pattern seen in the training data. The straight line fit associated with a linear regression model without polynomial terms would have done an abysmal job on such a nonlinear data set. Adding additional complexity allowed it to better conform to the training set data. 
```

chapters/overfitting.qmd
```diff
@@ -0,0 +1,487 @@
...
+
+The problem is that added complexity can also add risk. Being more capable of adapting to subtle trends in the training data is only a good thing when such trends _generalize_ to other data. Recall in @sec-effect-encodings where we were estimating the effect of an agent on the price of a hotel room. One agent had a sinlge booking. The naive estimate of the ADR wouldn't generalize well. As they accrue more and more bookings, their mean would undoubtedly change to a more realistic value. Using the raw ADR value would "overfit to the training set data."
```
kjell-stattenacity, line 59:
```suggestion
The problem is that added complexity can also add risk. Being more capable of adapting to subtle trends in the training data is only a good thing when such trends _generalize_ to other data. Recall in @sec-effect-encodings where we were estimating the effect of an agent on the price of a hotel room. One agent had a single booking. The naive estimate of the ADR wouldn't generalize well. As they accrue more and more bookings, their mean would undoubtedly change to a more realistic value. Using the raw ADR value would "overfit to the training set data."
```

chapters/overfitting.qmd
```diff
@@ -0,0 +1,487 @@
...
+
+As an example, @fig-two-class-overfit shows a scenario where there are two predictors (A and B) and each point belongs to one of two classes (vermilion circles or blue triangles). These data were simulated and panel (a) shows a thick grey line representing the true class boundary where there is a 50% chance that a data point belongs to either class. Points to the left of the grey curve are more likely to be circles than points to the right. The curve itself has a smooth, parabolic shape. 
+
+We can fit a model to these data and estimate the class boundary. The model we used^[A support vector machine. See @sec-cls-svm.] has a tuning parameter call the _cost value_. For now, you can think of this parameter as one that discourages the model from making an incorrect prediction (i.e., placing it on the wrong side of the estimated boundary). In other words, as the cost parameter increases, the model is more and more incentivized to be accurate on the training data.
```
kjell-stattenacity, line 65:
```suggestion
We can fit a model to these data and estimate the class boundary. The model we used^[A support vector machine. See @sec-cls-svm.] has a tuning parameter called the _cost value_. For now, you can think of this parameter as one that discourages the model from making an incorrect prediction (i.e., placing it on the wrong side of the estimated boundary). In other words, as the cost parameter increases, the model is more and more incentivized to be accurate on the training data. 
```

chapters/overfitting.qmd
```diff
@@ -0,0 +1,487 @@
...
+    linewidth = 1 / 2,
+    alpha = 3 / 4
+  )  +
+  ggtitle("(b) Low complexity") +
+  labs(x = "Predictor A", y = NULL)
+
+base_plot + low_plot + mid_plot + high_plot + plot_layout(nrow = 1, byrow = FALSE)
+```
+
+When the cost is increased, @fig-two-class-overfit(c) shows an increase in complexity. The boundary is more parabolic, and a few more training set points are on the correct side of the estimated line.
```
kjell-stattenacity, line 194:
```suggestion
When the cost is increased, the complexity also increases (@fig-two-class-overfit(c)). The boundary is more parabolic, and a few more training set points are on the correct side of the estimated line. 
```

chapters/overfitting.qmd
```diff
@@ -0,0 +1,487 @@
...
+
+Another sign that a model probably has too much complexity is the additional component of the decision boundary at the bottom of @fig-two-class-overfit(d) (red arrow). Points inside this part of the boundary would be classified as triangles even though no triangles are near it. This is a sort of "blowback" of an overly complex model where choices the model makes in one part of the predictor space add complexity in a completely different part of the space. This blowback is often seen in areas where there are no data. It’s unnecessary and points to the idea that the model is trying to do too much. 
+
+This boundary demonstrates how added complexity can be a bad thing. A test set of data was simulated and is shown in @fig-two-class-new. The performance for this overly complex model is poor since the extra areas induced by the high cost value do not contain the right class of points.
```
kjell-stattenacity, line 200:
```suggestion
This boundary demonstrates how added complexity can be a bad thing. A test set of data was simulated and is shown in @fig-two-class-new. The performance for this overly complex model is far from optimal since the extra areas induced by the high cost value do not contain the right class of points. 
```

chapters/overfitting.qmd
```diff
@@ -0,0 +1,487 @@
...
+	
+## The Ways Models Overfit {#sec-how-to-overfit}
+
+This leads us to a formal definition of overfitting: 
+
+::: {.important-box}
+
+Overfitting is the situation where a model over-interprets trends and patterns in the training set that do not generalize to new data. 
+
+These irreproducible data trends result in models that **only do well on the training set**.
```
kjell-stattenacity, line 252:
```suggestion
These irreproducible data trends result in models that **do very well on the training set** but greatly underperform on the validation set. 
```

chapters/overfitting.qmd
```diff
@@ -0,0 +1,487 @@
...
+
+Suppose some statistical inferential method, such as a t-test, is used to determine if a predictor can differentiate between two classes. Small sample sizes make the chance of uninformative predictors falsely passing the filter large. If the same data are then used to build and evaluate the model, there is a significant probability that an ineffective model will appear to be very predictive [@Ambroise2002p1493;@fes]. Again, using a data set external to the model training and development process shows you when this occurs. 
+
+Finally, it is possible to overfit a model via sample filtering. For example, in some machine learning competitions, participants might sub-sample the training set to be as similar to the unlabelled test set as possible. This will most likely improve their test set performance, but it generally handicaps a model for other data sets.
```
kjell-stattenacity, line 262:
```suggestion
Finally, it is possible to overfit a model via sample filtering. For example, in some machine learning competitions, participants might sub-sample the training set to be as similar to the unlabelled test set as possible. This will most likely improve their test set performance, but it generally handicaps a model when predicting a new yet-to-be-seen data set. 
```

chapters/overfitting.qmd
```diff
@@ -0,0 +1,487 @@
...
+
+Finally, it is possible to overfit a model via sample filtering. For example, in some machine learning competitions, participants might sub-sample the training set to be as similar to the unlabelled test set as possible. This will most likely improve their test set performance, but it generally handicaps a model for other data sets. 
+
+## External Data to Measure Effectiveness {#sec-external-validation}
+
+By now, it should seem clear that the best way of avoiding overfitting is to quantify the model using a separate data set. How do we do that? We've said that most modeling projects create an initial partition of the data into training and testing sets. Can the testing set detect overfitting?
```
kjell-stattenacity, line 266:
```suggestion
By now, it should seem clear that the best way of avoiding overfitting is to quantify the model performance using a separate data set. How do we do that? We've said that most modeling projects create an initial partition of the data into training and testing sets. Can the testing set detect overfitting? 
```

chapters/overfitting.qmd
```diff
@@ -0,0 +1,487 @@
...
+
+We’ll again use the classification data from @fig-two-class-overfit to demonstrate, this time with a different model. A boosted tree, described in more detail later, is a collection of individual decision trees created sequentially. Each depends on the previous trees. For a new sample, each tree is predicted, and a weighted average is used to create the predicted class probabilities. We’ll use 100 trees to create our ensemble^[But, obviously, this is yet another tuning parameter.]. However, we don’t know how deep they should be. One method of controlling the number of splits in a tree is to control how much data should be in a partition to continue splitting. For example, in @fig-reg-tree, the left-most terminal node in the tree contained $n_{tr}$ = 971 data points after all splits are applied. If we just keep splitting with no constraint, at some point, the amount of data in the node is too small to continue.
```
kjell-stattenacity, line 268:
```suggestion
We’ll again use the classification data from @fig-two-class-overfit to demonstrate, this time with a different model. A boosted tree, described in more detail later, is a collection of individual decision trees created sequentially. Each depends on the previous trees. For a new sample, each tree is used to predict the outcome, and a weighted average is used to create the predicted class probabilities. We’ll use 100 trees to create our ensemble^[But, obviously, this is yet another tuning parameter.]. However, we don’t know how deep the trees should be. One method of controlling the number of splits in a tree is to control how much data should be in a partition to continue splitting. For example, in @fig-reg-tree, the left-most terminal node in the tree contained $n_{tr}$ = 971 data points after all splits are applied. If we just keep splitting with no constraint, at some point, the amount of data in the node is too small to continue. 
```

chapters/overfitting.qmd
```diff
@@ -0,0 +1,487 @@
...
+  scale_color_manual(values = method_cols)
+```	
+
+We suggest using either of these two approaches (validation set or resampling) to measure performance during model development. Before going to the next chapter, let's give a preview of [Chapters -@sec-grid-search] and [-@sec-iterative-search].
+
+## How Should We Optimize Complexity? {#sec-optimizing-complexity}
+
+Now that we know how to appropriately estimate metrics for evaluating effectiveness, how do we choose tuning parameter values? Do we pick them at random? 
+
+Operationally, there are two main classes of search routines for optimizing model complexity via tuning parameters. The oldest is _grid search_. For each tuning parameter in the model, we define a reasonable set of specific candidate values to investigate. When there is more than one tuning parameter, there are various ways to create a multidimensional grid (these are discussed in @sec-grid-search). Each candidate in the grid is evaluated and the "best" combination of parameter values is chosen as the ones to use (if we think the model works well enough).
```
kjell-stattenacity, line 422:
```suggestion
Operationally, there are two main classes of search routines for optimizing model complexity via tuning parameters. The oldest is _grid search_. For each tuning parameter in the model, we define a reasonable set of specific candidate values to investigate. When there is more than one tuning parameter, there are various ways to create a multidimensional grid (these are discussed in @sec-grid-search). Each candidate in the grid is evaluated and the "best" combination of parameter values is chosen as the one to use (if we think the model works well enough). 
```

chapters/interactions-nonlinear.qmd
```diff
@@ -789,6 +789,196 @@ The H-statistic and its alternatives can be an incredibly valuable tool for lear
...
+
+Bias is the difference between some estimate, like an average or a model prediction, and its true value. The true value is not the same as the data we collect; it is the unknowable theoretical value. For this reason, bias is often difficult to compute directly. For machine learning, the most pertinent idea of bias relates to how well a model can conform to the patterns we see in the data (hoping that the observed data is a good representation of the true values). Back in @fig-global-polynomial, we saw that a linear model didn’t fit the data well. This is due to simple linear regression being a high-bias model because, without additional feature engineering, it cannot replicate nonlinear trends. Adding polynomial or spline terms decreased the model’s bias since it was more flexible.
+
+::: {.important-box}
+We can think of bias in terms of the question: "How close are we _aiming_ at the target?" and variance as: "How much do our results vary when shooting at the target?"
```
kjell-stattenacity, line 802:
```suggestion
We can think of bias in terms of the question: "How close are we _aiming_ to the center of the target?" and variance as: "How much do our results vary when shooting at the target?"
```

chapters/interactions-nonlinear.qmd
```diff
@@ -789,6 +789,196 @@ The H-statistic and its alternatives can be an incredibly valuable tool for lear
...
+:::
+
+The concepts of variance and bias are paired because they are often at odds with one another. For models that predict, we’d like a _low variance_, _low bias_ model. In many cases, that can be exceedingly difficult. We often lower bias by adding model complexity. However, increased complexity usually comes at the expense of stability (i.e., high variance). We will often be in the position of trying to find an acceptable compromise between the two. This was discussed with global polynomials; linear models were bad for our data, but adding too many polynomial terms adds unnecessary complexity and an explosion of variance (reflected in the confidence bands).
```
kjell-stattenacity, line 805:
```suggestion
The concepts of variance and bias are paired because they are often at odds with one another. For models that predict, we’d like a _low variance_, _low bias_ model. In many cases, that can be exceedingly difficult to achieve. We can often lower bias by adding model complexity. However, increased complexity usually comes at the expense of stability (i.e., high variance). We will often be in the position of trying to find an acceptable compromise between the two. This was discussed with global polynomials; linear models were bad for our data, but adding too many polynomial terms adds unnecessary complexity and an explosion of variance (reflected in the confidence bands). 
```

chapters/interactions-nonlinear.qmd
```diff
@@ -789,6 +789,196 @@ The H-statistic and its alternatives can be an incredibly valuable tool for lear
...
+  te$degree <- deg
+  te
+}
+sim <- function(deg, n = 30) {
+  make_data(num_points = n) %>% 
+    get_preds(deg)
+}
+```
+
+To create a simulated data set, `r nrow(make_data())` samples were simulated. Of these, `r nrow(make_data()) - 1` were used for the training set and one were reserved for the demonstrating the variance and bias. The test set values were roughly equally spaced across the range of the predictor. One simulated data set is shown in @fig-nonlinear-sim, along with the true underlying pattern.
```
kjell-stattenacity, line 858:
```suggestion
To create a simulated data set, `r nrow(make_data())` samples were generated. Of these, `r nrow(make_data()) - 1` were used for the training set and one was reserved for the estimating the variance and bias. The test set values were roughly equally spaced across the range of the predictor. One simulated data set is shown in @fig-nonlinear-sim, along with the true underlying pattern.  
```

chapters/interactions-nonlinear.qmd
```diff
@@ -789,6 +789,196 @@ The H-statistic and its alternatives can be an incredibly valuable tool for lear
...
+    .by = c(degree, point)
+  ) %>% 
+  mutate(
+    double_check = var + bias^2, 
+    mse = rmse^2, 
+    point = format(point)
+  )
+```
+
+A linear model was estimated using ordinary least squared and a polynomial expansion for each simulated data set. The test set values were predicted, and the bias, variance, and root mean squared error statistics were computed. This was repeated 50,000 times for polynomial degrees ranging from one to twenty.
```
kjell-stattenacity, line 901:
```suggestion
A linear model was estimated using ordinary least squares and a polynomial expansion for each simulated data set. The test set values were predicted, and the bias, variance, and root mean squared error statistics were computed. This was repeated 50,000 times for polynomial degrees ranging from one to twenty.  
```

chapters/interactions-nonlinear.qmd
```diff
@@ -789,6 +789,196 @@ The H-statistic and its alternatives can be an incredibly valuable tool for lear
...
+
+@fig-sim-results shows the average statistic values across model complexity (i.e., polynomial degree). A linear model performs poorly for the bias due to underfitting (as expected). In panel (a), Adding more nonlinearity results in a substantial decrease in bias because the model fit is closer to the true equation. This improvement plateaus at a sixth-degree polynomial and stays low (nearly zero). The variance begins low; even though the linear model is ineffective, it is stable. Once additional terms are added, the variance of the model steadily increases then explodes around an 20<sup>th</sup> degree polynomial^[In this particular case, the variance becomes very large since the number of parameters is nearly the same as the number of training set points (`r nrow(make_data()) - 1`). This makes the underlying mathematical operation (matrix inversion) numerically unstable. Even so, it is the result of excessive model complexity.]. This shows the tradeoff; as the model becomes more complex, it fits the data better but eventually becomes unstable.
```
kjell-stattenacity, line 903:
```suggestion
@fig-sim-results shows the average statistic values across model complexity (i.e., polynomial degree). A linear model performs poorly for the bias due to underfitting (as expected). In panel (a), adding more nonlinearity results in a substantial decrease in bias because the model fit is closer to the true equation. This improvement plateaus at a sixth-degree polynomial and stays low (nearly zero). The variance begins low; even though the linear model is ineffective, it is stable. Once additional terms are added, the variance of the model steadily increases then explodes around an 20<sup>th</sup> degree polynomial^[In this particular case, the variance becomes very large since the number of parameters is nearly the same as the number of training set points (`r nrow(make_data()) - 1`). This makes the underlying mathematical operation (matrix inversion) numerically unstable. Even so, it is the result of excessive model complexity.]. This shows the tradeoff; as the model becomes more complex, it fits the data better but eventually becomes unstable.
```

chapters/interactions-nonlinear.qmd
```diff
@@ -789,6 +789,196 @@ The H-statistic and its alternatives can be an incredibly valuable tool for lear
...
+
+@fig-tradeoff shows a simple example. Suppose we have some alternative method for estimating the mean of a group that adds some bias while reducing variance. That tradeoff might be worthwhile if some bias will greatly reduce the variance. In our figure, suppose the true population mean is zero. The green curve represents the sample mean for a fixed sample size. It is centered at zero but has significant uncertainty. The other curve might be an alternative estimator that produces a slightly pessimistic estimate (its mean is slightly smaller than zero) but has 3-fold smaller variation. This may be a good idea depending on how the estimate will be used ^[This specific example will be referenced again when discussing resampling in @sec-resampling.].
```
kjell-stattenacity, line 957:
```suggestion
@fig-tradeoff shows a simple example. Suppose we have some alternative method for estimating the mean of a group that adds some bias while reducing variance. That tradeoff might be worthwhile if some bias will greatly reduce the variance. In our figure, suppose the true population mean is zero. The green curve represents the sample mean for a fixed sample size. It is centered at zero but has significant uncertainty. The other curve might be an alternative estimator that produces a slightly pessimistic estimate (its mean is slightly smaller than zero) but has 3-fold smaller variation. When estimating the location of the mean of the population, the biased estimate will have a smaller confidence interval for a given sample size than the unbiased estimate.  While the biased estimate may slightly miss the target, the window of the location will be much smaller than the unbiased estimate.  This may be a good idea depending on how the estimate will be used ^[This specific example will be referenced again when discussing resampling in @sec-resampling.].
```

chapters/whole-game.qmd
```diff
@@ -619,7 +619,7 @@ nnet_cap <-
...
...odel). Other details: each model could be trained for up to `r nnet_args$epochs` epochs. However, a small amount of the training set was randomly set aside to monitor the sums of squared errors at each epoch. Early stopping was triggered when this out-of-sample loss degraded for `r nnet_args$stop_iter` consecutive epochs. A rectified linear unit activation function was used between the predictor and hidden layers. A `r nnet_args$rate_schedule` learning rate scheduler was used with a maximum rate of `r nnet_args$learn_rate` and a weight decay value of `r nnet_args$penalty` was applied. In a more complete analysis, all of these values would also be tuned.]? This decision will determine the complexity of the model. As models become model complex, they have more capacity to find complicated relationships between the predictors and response. However, added complexity adds to the risk of overfitting. This situation, discussed in @sec-complexity, occurs when a model fits exceptionally well to the training data but fails for new data when it finds patterns in the training data that are artifacts (i.e., do not generalize for other data sets). How should we choose the number of hidden units?
```
kjell-stattenacity, line 622:
```suggestion
One question about this model is: how many hidden units should we use^[This model description is deliberately vague (so as not to confuse readers who are new to this model). Other details: each model could be trained for up to `r nnet_args$epochs` epochs. However, a small amount of the training set was randomly set aside to monitor the sums of squared errors at each epoch. Early stopping was triggered when this out-of-sample loss degraded for `r nnet_args$stop_iter` consecutive epochs. A rectified linear unit activation function was used between the predictor and hidden layers. A `r nnet_args$rate_schedule` learning rate scheduler was used with a maximum rate of `r nnet_args$learn_rate` and a weight decay value of `r nnet_args$penalty` was applied. In a more complete analysis, all of these values would also be tuned.]? This decision will determine the complexity of the model. As models become model complex, they have more capacity to find complicated relationships between the predictors and response. However, added complexity increases the risk of overfitting. This situation, discussed in @sec-complexity, occurs when a model fits exceptionally well to the training data but fails for new data when it finds patterns in the training data that are artifacts (i.e., do not generalize for other data sets). How should we choose the number of hidden units?
```

kjell-stattenacity (changes requested):

PR #49, opened 2024-08-26
Resampling (chunk 2/3)

topepo:
Adds other basic resampling methods. 

Next chunk is time series and spatial CV plus FAQ section. 

I might re-do some of the initial splitting materials for spatial resampling. We might have a good classification data set on a [forestry](https://github.com/simonpcouch/forested) application.


chapters/resampling.qmd
```diff
@@ -263,6 +261,327 @@ mc_bias + mc_prec +
...
+
+
+### Repeated Cross-Validation {.unnumbered}
+
+One way to further improve the precision is to use **repeated V-fold cross-validation**. In this case, the same sample allocation scheme is used to create the folds and this is repeated multiple times using different random number seeds. For example, two repeats of 10-fold cross-validation^[This isn’t the same as using _V_ = 20. That scheme has very different bias properties than two repetitions of _V_ = 10.] create two sets of _V_ folds, which are treated as a collection of twenty resamples (e.g., _B_ = _V_ $\times$ _R_ for _R_ repeats). Again, increasing the number of resamples does not generally change the bias.
```
kjell-stattenacity, line 285:
```suggestion
One way to further improve the precision is to use **repeated V-fold cross-validation**. In this case, _V_-fold CV is repeated multiple times using different random number seeds.  For example, two repeats of 10-fold cross-validation^[This isn’t the same as using _V_ = 20. That scheme has very different bias properties than two repetitions of _V_ = 10.] create two sets of _V_ folds, which are treated as a collection of twenty resamples (e.g., _B_ = _V_ $\times$ _R_ for _R_ repeats). Again, increasing the number of resamples does not generally change the bias. 
```

chapters/resampling.qmd
```diff
@@ -263,6 +261,327 @@ mc_bias + mc_prec +
...
+  geom_hline(yintercept = 0, lty = 2) +
+  labs(y = "Std. Error", x = "Number of Resamples")  +
+  scale_color_brewer(palette = "Dark2")
+
+vfold_bias + vfold_prec +
+  plot_layout(guides = 'collect') &
+  theme(legend.position = "top")
+```
+
+Is there any advantage to using _V_ > 10? Not much. The decrease in bias is small and many more replicates are required to reach the same precision as _V_ = 10.  For example, twice as many resamples is required for 20-fold CV to match the variance of 10-fold CV.
```
kjell-stattenacity, line 352:
```suggestion
Is there any advantage to using _V_ > 10? Not much. The decrease in bias is small and many more replicates are required to reach the same precision as _V_ = 10.  For example, twice as many resamples are required for 20-fold CV to match the variance of 10-fold CV. 
```

chapters/resampling.qmd
```diff
@@ -263,6 +261,327 @@ mc_bias + mc_prec +
...
+cv_10_bias <- 
+  median(
+    stats_v_fold_cv$bias[
+      stats_v_fold_cv$model == "logistic" & 
+        stats_v_fold_cv$folds == "10" & 
+        stats_v_fold_cv$.metric == "brier_class"]) * 100
+```
+
+
+@fig-cv-stats shows bias and variance results from the simulation, where the x-axis is the total number of resamples $V\times R$. As expected, the bias decreases with _V_ and is constant over the number of resamples. Five-fold cross-validation stands out as particularly bad with a percent bias of roughly `r signif(cv_5_bias, 2)`%. Using ten folds decreases this to about `r signif(cv_10_bias, 2)`%. The precision values show that smaller values of _V_ have better precision; 10-fold CV needs substantially more resamples to have the same standard error than _V_ = 5..
```
kjell-stattenacity, line 314:
```suggestion
@fig-cv-stats shows bias and variance results from the simulation, where the x-axis is the total number of resamples $V\times R$. As expected, the bias decreases with _V_ and is constant over the number of resamples. Five-fold cross-validation stands out as particularly bad with a percent bias of roughly `r signif(cv_5_bias, 2)`%. Using ten folds decreases this to about `r signif(cv_10_bias, 2)`%. The precision values show that smaller values of _V_ have better precision; 10-fold CV needs substantially more resamples to have the same standard error than _V_ = 5.
```

chapters/resampling.qmd
```diff
@@ -263,6 +261,327 @@ mc_bias + mc_prec +
...
+vfold_bias + vfold_prec +
+  plot_layout(guides = 'collect') &
+  theme(legend.position = "top")
+```
+
+Is there any advantage to using _V_ > 10? Not much. The decrease in bias is small and many more replicates are required to reach the same precision as _V_ = 10.  For example, twice as many resamples is required for 20-fold CV to match the variance of 10-fold CV. 
+
+For this simulation, the properties are about the same when MCCV and 10-fold CV are matched in terms of number of resamples and the amount allocated to the assessment set. 
+
+Our recommendation is to almost always use _V_ = 10. The bias and variance gains are both good with 10 folds. Reducing bias in 5-fold cross-validation is difficult but the 10-fold precision can be improved by increased replication.
```
kjell-stattenacity, line 356:
```suggestion
Our recommendation is to almost always use _V_ = 10. The bias and variance improvements are both good with 10 folds. Reducing bias in 5-fold cross-validation is difficult but the precision for 10-fold can be improved by increased replication. 
```

chapters/resampling.qmd
```diff
@@ -263,6 +261,327 @@ mc_bias + mc_prec +
...
+
+## The Bootstrap {#sec-bootstrap}
+
+The bootstrap [@Efron1979boot;@davison1997bootstrap;@efron2003second;@EfronHastie2016a] is a resampling methodology originally created to compute the sampling distribution of statistics using minimal probabilistic assumptions^[This application will be discussed in section TODO.]. In this chapter, we'll define a bootstrap sample and show how it can be used to measure fit quality for predictive models. 
+
+For a training set with $n_{tr}$ data points, a bootstrap resample takes a random sample of the training set that is also size $n_{tr}$. It does this by sampling with replacement; when each of the $n_{tr}$ samples is drawn, it has no memory of the prior selections. This means that each row can be randomly selected again. For example, @fig-bootstrap-scheme shows another schematic for three bootstrap samples. In this figure, the thirteenth training set sample was selected to go into the first analysis set three times. For this reason, a bootstrap resample will contain multiple replicates of some training set points, while others will not be selected at all. The data that were never selected are used to create the assessment set.
```
kjell-stattenacity, line 362:
```suggestion
For a training set with $n_{tr}$ data points, a bootstrap resample takes a random sample of the training set that is also size $n_{tr}$. It does this by sampling with replacement; when each of the $n_{tr}$ samples is drawn, it has no memory of the prior selections. This means that each row can be randomly selected again. For example, @fig-bootstrap-scheme shows another schematic for three bootstrap samples. In this figure, the thirteenth training set sample was selected to go into the first analysis set three times. For this reason, a bootstrap resample will contain multiple replicates of some training set points, while others will not be selected at all. The data that were never selected are used to create the assessment set.  This means that while the training set will have the same number of data points, the assessment sets will have varying numbers of data points.
```

chapters/resampling.qmd
```diff
@@ -263,6 +261,327 @@ mc_bias + mc_prec +
...
+#| out-width: '70%'
+#| fig-cap: "A schematic of three bootstraps resamples created from an initial pool of 30 data points."
+knitr::include_graphics("../premade/bootstraps.svg")
+```
+
+The probability that a data point will be picked is $1 / n_{tr}$ and, from this, the probability that a training set point is _not selected at all_ is 
+
+$$\prod_{i=1}^{n_{tr}} \left(1 - \frac{1}{n_{tr}}\right) = \left(1 - \frac{1}{n_{tr}}\right)^{n_{tr}}\approx e^{-1} = 0.368$$
+
+Since the bootstrap sample contains the _selected_ data, each training point is selected with probability $1 - e^{-1} \approx `r round((1 - exp(-1)), 3)`$. The implication is that, on average, the analysis set, contains about `r round((1 - exp(-1)) * 100, 1)`% unique training set points and the assessment set includes, on average, 36.8% of the training data. If we were to compare the amount of unique training set points held out by the bootstrap to V-fold cross-validation, the bootstrap is comparable to using $V = 3$.
```
kjell-stattenacity, line 378:
```suggestion
Since the bootstrap sample contains the _selected_ data, each training point is selected with probability $1 - e^{-1} \approx `r round((1 - exp(-1)), 3)`$. The implication is that, on average, the analysis set, contains about `r round((1 - exp(-1)) * 100, 1)`% unique training set points and the assessment set includes, on average, 36.8% of the training data.  When comparing the number of unique training set points excluded by the bootstrap method to those excluded by V-fold cross-validation, the bootstrap method is roughly equivalent to using $V=3$ in cross-validation. 
```

kjell-stattenacity (changes requested):

PR #53, opened 2024-09-03
Spatial forested example

topepo:
Remove the old spatial example and use @simonpcouch's forested package with hexagonal blocks. 

(To review, you might comment out the other `qmd` files from `_quarto.yml` so that it runs much faster)

Perhaps @mikemahoney218 will have feedback on this too.


R/setup_forested.R
```diff
@@ -0,0 +1,129 @@
...
+  st_as_sf(coords = c("lon","lat"), crs = st_crs("EPSG:4326"))
+
+
+# ------------------------------------------------------------------------------
+# Conduct the initial split using block sampling and buffering
+
+set.seed(318)
+forested_split <-
+  spatial_initial_split(forested_sf, prop = 1 / 5, spatial_block_cv,
+                        buffer = 80 * 80, method = "continuous",
```
topepo, line 40: Does anyone know what the units are here? Meters per degree?

R/setup_forested.R
```diff
@@ -0,0 +1,129 @@
...
+  buffer = 80 * 80,
+  method = "continuous",
+  n = 25,
+  square = FALSE
+)
+
+# ------------------------------------------------------------------------------
+# re-calculate geocodes and strip off geometry column
+
+re_geocode <- function(x) {
```
topepo, line 78: We'll be using this data in later chapters and will use standard ML tools. Since the `geometry` is sticky, we delete it from the data/resamples and add the longitude and latitude back in. 

Perhaps there is an easier way to do this>

R/setup_forested.R
```diff
@@ -0,0 +1,129 @@
...
+# These computations take a while so we do them in batch mode then load the 
+# results in later chapters
+
+# ------------------------------------------------------------------------------
+# We remove some of the existing algorithmic data already there. 
+
+for_analysis <-
+  forested %>%
+  rename(class = forested) %>%
+  select(-tree_no_tree, -land_type, -canopy_cover)
```
topepo, line 24: These are algorithmically derived features. Building on models with them seems a little redundant. Also, we want to make the problem a little more difficult.

DESCRIPTION
```diff
@@ -123,7 +125,8 @@ Suggests:
     testthat (>= 3.0.0)
 Remotes:
     Bioconductor/BiocParallel,
-    mixOmicsTeam/mixOmics
+    mixOmicsTeam/mixOmics,
+    simonpcouch/forested
```
simonpcouch, line 129: I _think_ the CRAN version is up to date! We don't plan on switching out the data to different states in that package, for now.

chapters/initial-data-splitting.qmd
```diff
@@ -362,15 +380,102 @@ When splitting multi-level data into a training and test set, the data are split
...
+ - Buffering [@brenning2005spatial;@le2014spatial] is a method that excludes data around a specific data point or group. 
+ 
+We'll use both of these methods below for illustration. @Pohjankukka2017, @lyons2018comparison, @meyer2019importance @karasiak2022spatial, and @mahoney2023assessing have good overviews of the importance of countering spatial autocorrelation and appropriate sampling methods. 
+
+For our example, we’ll use blocking and buffering to allocate roughly 20% of the data to the testing set. First, the locations were allocated to a 25 x 25 grid of hexagons across the state. After this, a buffer of `r format(80^2, big.mark = ",")` squared map units was created around the hexagon such that outside points within the buffer would be excluded from being used in the training or testing sets. Finally, roughly 20% of the locations were selected for the test set by moving through the space and selecting hexagons sequentially. @fig-forested-split shows the results of this process.
```
simonpcouch, line 449: "Blocking" is not a word that had yet appeared in this subsection. Does this refer to the first bullet point above?

simonpcouch: A switch from present and future tenses to past tense happens at "First, ..." and switches back after two paragraphs.

chapters/initial-data-splitting.qmd
```diff
@@ -362,15 +380,102 @@ When splitting multi-level data into a training and test set, the data are split
...
 
-Validation sets serve the same purpose as resampling methods described in @sec-validation and we can consider them single resamples of the training data. Methods like bootstrapping or cross-validation use many alternative versions of the training set to compute performance statistics. When our data are extensive, multiple resamples are computationally expensive without significantly improving the precision of our estimates. 
+> "Everything is related to everything else, but near things are more related than distant things."
 
-Without loss of generalization, we will treat the validation set as a particular case of resampling where there is a single resample of the training set. This difference is not substantive and allows us to have a common framework for measuring model efficacy (before the testing set).
+Think of an overhead color satellite image of a forest. Different types of trees have different foliage colors. If we focus on the color green, the distribution of "greeness" is not a random pattern of the x/y axes (random distribution would resemble static). In other words, pixels cannot be assumed to be independent of one another.
```
simonpcouch, line 389:
```suggestion
Think of an overhead color satellite image of a forest. Different types of foliage—or lack thereof—have different colors. If we focus on the color green, the distribution of "greeness" is not a random pattern of the x/y axes (random distribution would resemble static). In other words, pixels cannot be assumed to be independent of one another. 
```

chapters/initial-data-splitting.qmd
```diff
@@ -362,15 +380,102 @@ When splitting multi-level data into a training and test set, the data are split
...
 
-## Validation Sets  {#sec-three-way-split}
+## Splitting Spatial Data {#sec-spatial-splitting}
 
-As previously discussed, validation sets are a separate partition of the data that function as a precursor for the testing set. It allows us to obtain performance estimates on our model(s) during the development cycle. These are commonly used in deep learning and other domains where the initial data sizes range from very large to massive. This additional partition is often created simultaneously with the training and testing sets. 
+Spatial data are largely about the location of data points. As previously mentioned, the Ames data can be considered spatial since each row of the data set corresponds to a property at a specific location. Spatial data can have _autocorrelation_, with objects being more similar to closer objects than to objects further away. Tobler's First Law of Geography [@bjorholm2008extent] is:
```
simonpcouch, line 385: First sentence feels quite vague.

simonpcouch (approved):
Super solid.

I asked Grayson for a citation on the spatial decay of the outcome(s) yesterday—will drop that here when he passes this along.

Reading this also brought up the question of what the original sampling design of these data are in the first place, as the lat/lons are clearly not randomly sampled from throughout the state. I'll also ask him for a citation there, but the gist is that Washington State (71,362 square miles) is divided into 7109 6000-acre hexagons, and a "plot" is sampled from within each of those hexagons. The plots are comprised of four circular subplots with 24ft radius, so ~7,250square feet in a plot, where there is one subplot centered at the given lat/lon and the remaining three are positioned in an equilateral triangle around the center one. A plot is "forested" if a forest "stand" intersects with any of the four subplots. They have some proper definition of a "stand," but a group of trees, loosely.

All that to say—I think they set up that sampling design so that they can apply some stronger assumptions when splitting data and making inference. That may be too in the weeds for these chapters.

An aside: this website _feels_ really nice. Visually pleasant, nice reactivity. Bravo on the styling. :)

simonpcouch:
For the GH archaelogists--

re: spatial decay of the response
* https://link.springer.com/article/10.1007/s10651-007-0049-5
* https://link.springer.com/article/10.1007/s13253-024-00600-6

re: sampling design, see the USDA's "The Enhanced Forest Inventory and Analysis Program—National Sampling Design and Estimation Procedures", Bechtold and Patterson, aka "The Green Book"

PR #65, opened 2025-01-19
Iterative search, part 1

topepo:
Sections on 

- gradient-based optimization
- simulated annealing
- genetic algorithms
- Bayesian models

Part 2 of 2 will be on Bayesian optimization, Gaussian processes, and a case study.


chapters/iterative-search.qmd
```diff
@@ -0,0 +1,1027 @@
...
+
+The most commonly tuned SVM parameter is independent of the different kernel functions: the _cost value_. This dictates how much a poorly predicted model should penalize the performance metric (when evaluated on the training set). Higher cost values will encourage the SVM to become increasingly more complex. We've previously seen the effect of modulating the SVM cost value in @fig-two-class-overfit. For small costs, the SVM classification model did not try too hard to classify samples correctly. When increased to an extreme, it significantly overfit the training set points.
+
+Just as in @sec-embeddings, the RMSE will be computed from a validation set and these statistics will be used to guide our efforts. 
+
+Let's define a wide  space for our two tuning parameters: cost will vary from 2<sup>-10</sup> to 2<sup>10</sup> and the scale factor^[Why do these two parameters use different bases? It’s mostly a convention. Both parameters have valid values that can range across many orders of magnitude. The change in cost tends to occur more slowly than the scaling factor, so a smaller base of 2 is often used.] is allowed to range from 10<sup>-10</sup> to 10<sup>-0.1</sup>.
```
topepo, line 80: I'm wondering if I should just change both of these to be log-10.

chapters/iterative-search.qmd
```diff
@@ -0,0 +1,1027 @@
...
+::: {.column width="15%"}
+:::
+
+::: {.column width="70%"}
+
+::: {#fig-sa-example}
+
+::: {.figure-content}
+
+```{shinylive-r}
```
topepo, line 528: Once these are merged, I will move these apps to GitHub in individual files so that this file is much simplier.

topepo:
Here's a pdf: 

[12-Iterative-Search-Part-1.pdf](https://github.com/user-attachments/files/18466807/12-Iterative-Search-Part-1.pdf)

chapters/iterative-search.qmd
```diff
@@ -0,0 +1,1027 @@
...
+ - We can accept the current value as "suboptimal" and base our next perturbation on it, or
+ - discard the current solution as if it did not occur. The next candidate point is a perturbation of the last "acceptable" point. 
+
+For our SVM example, suppose we start with a candidate where $x_0 = \left[log_2 (cost), log_{10} (scale)\right[$ was $x_0 = [-10, -0.1]$ and had an associated RMSE value of 6.10%. For the next candidate, a perturbation of this values is created, say $x_1 = [-7.11, -0.225]$. Suppose that the corresponding RMSE was measured at 5.97%. Since $x_1$ has a better performance metric it is automatically accepted and is used to create the next parameter. 
+
+However, suppose that the RMSE value for $x_1$ was 7.00%, meaning that the new candidate did worse than the initial one.  
+
+Simulated annealing generates a probability threshold for accepting the worse solution. Typically, the probability depends on two quantities: 
+
+* The difference in prior and current estimates of the objective function. If the new candidate is nearly as good as the current one, the probability of acceptance should be larger than that of a severely unsatisfactory candidate.
```
kjell-stattenacity, line 458:
```suggestion
* The difference between the current and previous objective function values. If the new candidate is nearly as good as the current solution, the probability of acceptance will be higher compared to a candidate that is significantly worse.```

File: chapters/iterative-search.qmd [hunk 4]
```diff
@@ -0,0 +1,1027 @@
...
+
+For our SVM example, suppose we start with a candidate where $x_0 = \left[log_2 (cost), log_{10} (scale)\right[$ was $x_0 = [-10, -0.1]$ and had an associated RMSE value of 6.10%. For the next candidate, a perturbation of this values is created, say $x_1 = [-7.11, -0.225]$. Suppose that the corresponding RMSE was measured at 5.97%. Since $x_1$ has a better performance metric it is automatically accepted and is used to create the next parameter. 
+
+However, suppose that the RMSE value for $x_1$ was 7.00%, meaning that the new candidate did worse than the initial one.  
+
+Simulated annealing generates a probability threshold for accepting the worse solution. Typically, the probability depends on two quantities: 
+
+* The difference in prior and current estimates of the objective function. If the new candidate is nearly as good as the current one, the probability of acceptance should be larger than that of a severely unsatisfactory candidate. 
+* The probability of acceptance should decrease over search iterations. An exponentially decaying function, called a "cooling schedule," often used.
```
kjell-stattenacity, line 459:
```suggestion
* The probability of acceptance should decrease over time as the search progresses. This is often achieved using an exponentially decaying function, known as the "cooling schedule."
```

chapters/iterative-search.qmd
```diff
@@ -0,0 +1,1027 @@
...
+
+An often used equation for the probability, assuming that smaller values are better, is 
+
+$$
+Pr[accept] = \exp\Bigl[-i\bigl(\hat{Q}(\boldsymbol{\theta}_{i}) - \hat{Q}(\boldsymbol{\theta}_{i-1})\bigr)\Bigr]
+$$
+
+where $i$ is the iteration number. To compare 6.1% versus 7.0%, the acceptance probability is `r round(exp(-(7-6.10)) * 100, 1)`%. To make the determination, a random uniform number is generated and is used to compare to a random number to decide on the next step. If this difference were to occur at a later iteration, say $i = 5$, the probability would drop to `r round(exp(-5*(7-6.10)) * 100, 1)`%.
```
kjell-stattenacity, line 467: The statement "a random uniform number is generated and is used to compare to a random number" doesn't make sense to me.

chapters/iterative-search.qmd
```diff
@@ -0,0 +1,1027 @@
...
+
+One small matter is related to the scale of the objective function. The difference in the exponent is very sensitive to scale. If, for example, instead of percentages we were to use the proportions 0.61 and 0.70, the probability of acceptance would change from `r round(exp(-(7-6.10)) * 100, 1)`% to `r round(exp(-(.7-.610)) * 100, 1)`%. One way to mitigate this issue is to use a normalized difference by dividing the raw difference by the previous objective function (i.e., (0.70-0.61) / 0.70). This is the approach used in the SA analyses here. 
+
+This process is continued until either a pre-defined number of iterations is reached or if there is no improvement by a certain number of iterations. The best result found at any time in the optimization is used as the final value; there is no real notion of "convergance" for this method. Also, as previously mentioned, a restart rule can stop simulated annealing from delving too far into suboptimal regions if no best results have occurred within some period of time. When the processes is restarted, can be continue from either the last best candidate point from previous iterations or at a random point in the parameter space.
```
kjell-stattenacity, line 471:
```suggestion
This process continues until either a pre-defined number of iterations is reached or there is no improvement after a certain number of iterations. The best result found during the optimization process is used as the final value, as there is no formal concept of "convergence" for this method. Additionally, as mentioned earlier, a restart rule can prevent simulated annealing from getting stuck in suboptimal regions if no better results have been found within a certain timeframe. When the process is restarted, it can either continue from the best candidate found in previous iterations or start from a random point in the parameter space.
```

chapters/iterative-search.qmd
```diff
@@ -0,0 +1,1027 @@
...
+
+How should the candidates be perturbed from iteration to iteration? When the tuning parameters are all numeric, we can create a random distance and angle from the current values. A similar process can be used for integers by "flooring" them to the nearest whole number. For qualitative tuning parameters, a random subset of parameters is chosen to change to a different value chosen at random. The amount of change should be large enough to search the parameters space and potentially get out of a local optimum.
+
+It is also important to conduct computations on the parameters in their potentially transformed space. This ensures that the entire range of the parameter values is treated equally. For example, we’ve been working with the values on their log (base 2) scale for the SVM cost parameter. When perturbing the values, we make sure to change them in the log space. This is true of all of the other search methods described in this chapter.
```
kjell-stattenacity, line 475:
```suggestion
It is important to perform computations on the parameters in their transformed space to ensure that the full range of possible values is treated equally. For example, when working with the SVM cost parameter, we use its log (base 2) scale. When perturbing the parameter values, we make sure to adjust them in the log space rather than the original scale. This approach applies to all other search methods discussed in this chapter.```

File: chapters/iterative-search.qmd [hunk 8]
```diff
@@ -0,0 +1,1027 @@
...
+:::
+
+:::
+
+::: {.column width="25%"}
+:::
+
+:::: 
+ 
+@fig-sa-example contains an animation of the results of the SA search. In the figure, the initial points are represented by open circles, and a grey diagonal line shows the ridge of values exists that corresponds to the best RMSE results.
```
kjell-stattenacity, line 515:
```suggestion
@fig-sa-example contains an animation of the results of the SA search. In the figure, the initial points are represented by open circles, and a grey diagonal line shows the ridge of values that corresponds to the best RMSE results. 
```

chapters/iterative-search.qmd
```diff
@@ -0,0 +1,1027 @@
...
+
+Simulated annealing has some attractive qualities. First, the process of creating a new candidate works for any type of parameter (e.g., real, integer, or qualitative). This isn’t the case for the two other iterative methods we’ll discuss. Also, the perturbation process is very fast; there is very little computational overhead to acquire the next objective function value. Also, since we are effectively simulating new candidate sets, we can potentially apply constraints to parameters or groups of parameters. For example, if a tuning parameter could only be odd integers, this would not be a significant problem for simulated annealing.
```
kjell-stattenacity, line 679:
```suggestion
Simulated annealing has several attractive qualities. First, the process of generating new candidates works with any type of parameter, whether real, integer, or qualitative. This is not true for the other two iterative methods we’ll discuss. Additionally, the perturbation process is very fast, meaning there is minimal computational overhead to compute the next objective function value. Since we are effectively generating new candidate sets, we can also apply constraints to individual parameters or groups of parameters. For example, if a tuning parameter is restricted to odd integers, this would not pose a significant problem for simulated annealing.
```

chapters/iterative-search.qmd
```diff
@@ -0,0 +1,1027 @@
...
+
+Another disadvantage is that a single candidate is processed at a time. If the training set size is not excessive, we could parallel process the multiple candidates simultaneously. We could make a batch of perturbations and pick the best value to keep or apply the probabilistic process of accepting a poor value. 
+
+This optimization took `r prettyunits::pretty_sec((initial_time[3] + sa_time[3]) / nrow(sa_history))` per candidate to execute for these data. 
+
+## Genetic Algorithms  {#sec-genetic-algo}
+
+Genetic algorithms (GAs)[@Mitchell1996AnIT;@eiben2015introduction] are an optimization method that mimics the process of evolution via natural selection. Our application (parameter tuning) isn't well suited to GAs; a standard GA search will use hundreds to millions of objective function evaluations. We will use a small search due to practical considerations. Despite this, genetic algorithms can often find a solution near the optimal value fairly quickly and, as previously stated, there is often a region of acceptable candidates. Finally, GAs are very unlikely to become stuck in a locally optimal space.
```
kjell-stattenacity, line 689:
```suggestion
Genetic algorithms (GAs) [@Mitchell1996AnIT; @eiben2015introduction] are an optimization method that mimics the process of evolution through natural selection. While GAs are not ideally suited for parameter tuning in our case—since a standard GA search typically requires hundreds to millions of objective function evaluations—we will limit the search to a smaller scale for practical reasons. Nevertheless, genetic algorithms can often find solutions near the optimal value fairly quickly. Additionally, as previously mentioned, there is usually a region of acceptable candidates. Finally, GAs are highly unlikely to become trapped in a locally optimal solution.
```

chapters/iterative-search.qmd
```diff
@@ -0,0 +1,1027 @@
...
+
+### Assembling Generations {#sec-ga-generations}
+
+Now that we know how to make new candidates, we can form new generations. We start by choosing the size. Normally, the minimum population size within a generation is at least 25-50 candidates. This may be infeasible depending on the computational cost of the model and the characteristics of the training set, such as size. We can use fewer points in the generation, but the risk is that we will never sample any acceptable results. In that case, combining the best results only results in more mediocre candidates. Increasing the mutation rate may help avoid this problem. For the first iteration, random sampling of the space is often used. Unsurprisingly, we strongly recommend that a space-filling design for the initial candidate set.
```
kjell-stattenacity, line 732:
```suggestion
Now that we know how to generate new candidates, we can form new generations. The first step is to determine the population size. Typically, the minimum population size within a generation is 25 to 50 candidates. However, this may be infeasible depending on the computational cost of the model and the characteristics of the training set, such as its size. While fewer candidates can be used, the risk is that we may fail to sample any acceptable results. In such cases, combining the best results will only produce more mediocre candidates. Increasing the mutation rate can help mitigate this issue. For the first iteration, random sampling of the parameter space is commonly used. It is highly recommended to use a space-filling design for the initial candidate set to ensure a more comprehensive exploration. 
```

chapters/iterative-search.qmd
```diff
@@ -0,0 +1,1027 @@
...
+  distinct(fitness, generation)
+
+is_same_as_gen_1 <- best_by_gen$fitness == best_by_gen$fitness[1]
+same_as_gen_1 <- sum(is_same_as_gen_1) - 1
+
+other_best <- round(unique(best_by_gen$fitness[!is_same_as_gen_1]), 3)
+other_best <- paste0(other_best, "%")
+```
+
+To illustrate genetic algorithms in two dimensions, a population size of 8 was used for 7 generations. These are _not_ good default values but were instead chosen to be at parity with the previous SA search and the optimization method used in the next section. The values of the two tuning parameters were kept as floating point values. The parental selection was conducted using sampling weights proportional to the RMSE values (smaller being better), a mutation rate of 10%, and a single candidate was retained at each generation via elitism.  The computations within a generation where computed in parallel. In parallel, this optimization took `r prettyunits::pretty_sec(ga_time[3] / nrow(ga_history))` per candidate (on average).  @fig-ga-example illustrates the results of the search.
```
kjell-stattenacity, line 758:
```suggestion
To illustrate genetic algorithms in two dimensions, a population size of 8 was used for 7 generations. These values are _not_ optimal defaults but were selected to align with the previous SA search and the optimization method discussed in the next section. The two tuning parameters were kept as floating-point values. Parental selection was based on sampling weights proportional to the RMSE values (with smaller values being better). The mutation rate was set at 10%, and elitism was applied by retaining the best candidate from each generation. All computations within a generation were performed in parallel. On average, the optimization took `r prettyunits::pretty_sec(ga_time[3] / nrow(ga_history))` per candidate.  @fig-ga-example illustrates the results of the search.   
```

kjell-stattenacity (changes requested):
I think this chapter looks great!  Mostly made wording changes to clarify the content.  Please feel free to disregard any changes.

chapters/iterative-search.qmd
topepo: I'll update that.

PR #71, opened 2025-03-13
Feature selection chapter

topepo:
Will go _after_ the chapter on comparing models.


chapters/feature-selection.qmd
```diff
@@ -0,0 +1,525 @@
...
+@fig-selection-sa shows the results, where each point is the average of the 10 RMSE values for the SA at each iteration. The results clearly improve as the selection process mostly begins to include important predictors and discards a few of the noise columns. After approximately 120 iterations, the solution stabilizes.  This appears to be a very straightforward search process that clearly indicates when to stop the SA search. 
+
+```{r}
+#| label: fig-selection-sa
+#| echo: false
+#| out-width: 60%
+#| fig-width: 5
+#| fig-height: 3.25
+#| fig-cap: The progress of using simulated annealing to select features for.a neural network. The dashed green line indicators the best RMSE value achievable for these simulated data.
+load("../RData/nnet_sa_res.RData")
```
topepo, line 295: Computed in `sim_sa.R`

topepo:
PDF version: [13  Feature Selection.pdf](https://github.com/user-attachments/files/19235533/13.Feature.Selection.pdf)

chapters/feature-selection.qmd
```diff
@@ -0,0 +1,525 @@
...
+  parallel_over = "everything",
+  save_workflow = TRUE
+)
+
+reg_metrics <- metric_set(mae, rmse, rsq)
+```
+
+In Chapters [-@sec-numeric-predictors] through [-@sec-interactions-nonlinear], we described different methods for adding new predictors to the model (e.g., interactions, spline columns, etc.). We can add any predictors (or functions of predictors) based on first principles or our intuition or contextual knowledge. 
+
+We are also motivated to remove unwanted or irrelevant predictors. This can decrease model complexity, training time, computer memory, the cost of acquiring unimportant predictors, and so on. For some models, the presence of non-informative predictors can _decrease_ performance.
```
EmilHvitfeldt, line 49: > For some models

maybe hint that we will show which models later in the chapter

chapters/feature-selection.qmd
```diff
@@ -0,0 +1,525 @@
...
+)
+
+reg_metrics <- metric_set(mae, rmse, rsq)
+```
+
+In Chapters [-@sec-numeric-predictors] through [-@sec-interactions-nonlinear], we described different methods for adding new predictors to the model (e.g., interactions, spline columns, etc.). We can add any predictors (or functions of predictors) based on first principles or our intuition or contextual knowledge. 
+
+We are also motivated to remove unwanted or irrelevant predictors. This can decrease model complexity, training time, computer memory, the cost of acquiring unimportant predictors, and so on. For some models, the presence of non-informative predictors can _decrease_ performance. 
+
+We may want to remove predictors based on their distributions. Unsupervised feature selection focuses on just the predictor data.
```
EmilHvitfeldt, line 51: This first sentence feels incomplete to me. but can't tell why

chapters/feature-selection.qmd
```diff
@@ -0,0 +1,525 @@
...
+```
+
+In Chapters [-@sec-numeric-predictors] through [-@sec-interactions-nonlinear], we described different methods for adding new predictors to the model (e.g., interactions, spline columns, etc.). We can add any predictors (or functions of predictors) based on first principles or our intuition or contextual knowledge. 
+
+We are also motivated to remove unwanted or irrelevant predictors. This can decrease model complexity, training time, computer memory, the cost of acquiring unimportant predictors, and so on. For some models, the presence of non-informative predictors can _decrease_ performance. 
+
+We may want to remove predictors based on their distributions. Unsupervised feature selection focuses on just the predictor data. 
+
+More often, though, we are interested in supervised feature selection^[Recall that we use the terms "predictors" and "features" interchangeably. We favor the former term, but "features" is more commonly used when discussing the filtering of columns.]: the removal of predictors that do not appear to have a relationship to the outcome. This implies that we have a statistic that can compute some measure of each predictor’s importance.
```
EmilHvitfeldt, line 53: > This implies that we have a statistic that can compute some measure of each predictor’s importance.

Does it tho? I feel like we don't **technically** have a importance measure in wrapper methods. At least not in the traditional sense.

we have a measure for each subset of predictors, but i don't think we have a natural meassure one for each predictor itself.

this feels very nit-picky

chapters/feature-selection.qmd
```diff
@@ -0,0 +1,525 @@
...
+
+While it is easy to add predictors to a model, it can be very difficult (and computationally expensive) to properly determine which predictors to remove. The reason comes back to the misuse (or over-use) of data. We’ve described the problem of using the same data to fit and evaluate a model. By naively re-predicting the training set, we can accidentally get overly optimistic performance statistics. To compensate for this, resampling repeatedly allocates some data for model fitting and other data for model evaluation. 
+
+For feature selection, we have the same problem when the same data are used to measure importance, filter the predictors, and then train the model on the smaller feature set. It is similar to repeatedly taking the same test and then measuring our knowledge using the grade for the last test repetition. It doesn’t accurately measure the quantity that is important to us. This is the reason that this chapter is situated here instead of in Part 2; understanding overfitting is crucial.
```
EmilHvitfeldt, line 57: > This is the reason that this chapter is situated here instead of in Part 2; **understanding overfitting is crucial.**

this might be an artifact of me reading out of order, but this last sub-sentence feels like it comes out of nowhere

chapters/feature-selection.qmd
```diff
@@ -0,0 +1,525 @@
...
+## The Cost of Irrelevant Predictors  {#sec-irrelevant-predictors}
+
+```{r}
+#| label: noise-sim
+#| echo: false
+
+load("../RData/noise_simulation_results.RData")
+```
+
+Does it matter if the model is given extra predictors that are unrelated to the outcome? It depends on the model. To quantify this, data were simulated using the equation described in @hooker2004discovering and also used in @sorokina2008detecting:
```
EmilHvitfeldt, line 70:
```suggestion
Does it matter if the model is given extra predictors that are unrelated to the outcome? It depends on the model. To quantify this, data were simulated using the equation described in @hooker2004discovering and used in @sorokina2008detecting: 
```

chapters/feature-selection.qmd
```diff
@@ -0,0 +1,525 @@
...
+
+ 1. Penalized linear regression, generalized additive models, and RuleFit add a penalty to their calculations, restricting the model parameters from becoming abnormally large unless the underlying data warrant a large coefficient. This effectively reduces the impact of extra predictors by keeping their corresponding model coefficient at or near zero. 
+
+ 2. MARS and Cubist selectively include predictors in their regression equations, only including those specifically selected as having value. 
+
+The other unaffected models are tree-based. As seen in @fig-reg-tree, these types of models make different rules by selecting predictors to _split_ the data. If a predictor was not used in any split, it is functionally independent of the outcome in the prediction equation. This provides some insensitivity to the presence of non-informative predictors, although there is slight degradation in RMSE as the number of columns increases.
```
EmilHvitfeldt, line 126: I think this section could help for a slight bit of reemphasis. It says "describe two clusters of models" then followed by 1., 2. and "The other". Maybe 3 clusters of models?

chapters/feature-selection.qmd
```diff
@@ -0,0 +1,525 @@
...
+
+ 1. Penalized linear regression, generalized additive models, and RuleFit add a penalty to their calculations, restricting the model parameters from becoming abnormally large unless the underlying data warrant a large coefficient. This effectively reduces the impact of extra predictors by keeping their corresponding model coefficient at or near zero. 
+
+ 2. MARS and Cubist selectively include predictors in their regression equations, only including those specifically selected as having value. 
+
+The other unaffected models are tree-based. As seen in @fig-reg-tree, these types of models make different rules by selecting predictors to _split_ the data. If a predictor was not used in any split, it is functionally independent of the outcome in the prediction equation. This provides some insensitivity to the presence of non-informative predictors, although there is slight degradation in RMSE as the number of columns increases. 
+
+In summary, irrelevant columns only slightly affect ML models that automatically select features. Models that do not have this attribute can be crippled under the same circumstances.
```
EmilHvitfeldt, line 128:
```suggestion
In summary, irrelevant columns minimally affect ML models that automatically select features. Models that do not have this attribute can be crippled under the same circumstances. 
```

doesn't have to be that word, but i think "only slightly" was hard to read

chapters/feature-selection.qmd
```diff
@@ -0,0 +1,525 @@
...
+In summary, irrelevant columns only slightly affect ML models that automatically select features. Models that do not have this attribute can be crippled under the same circumstances. 
+
+## Different Levels of Features {#sec-feature-levels}
+
+When we talk about selecting features, we should be clear about what level of computation we mean. There are often two levels: 
+
+- **Original predictors** are the unmodified version of the data. 
+- **Derived predictors** refer to the set of columns that are present after feature engineering. 
+
+For example, if our model requires all numeric inputs, a categorical predictor is often converted to binary indicators. A categorical column for the day of the week is the original predictor, while the binary indicators for individual days are the derived predictor. Similarly, the 550 measurement predictors in the barley data are the original predictors, and embedded values, such as the PCA components, are the derived predictors.
```
EmilHvitfeldt, line 137:
```suggestion
For example, if our model requires all numeric inputs, a categorical predictor is often converted to binary indicators. A date-time column is the original predictor, while the binary indicators for individual days of the week are the derived predictor. Similarly, the 550 measurement predictors in the barley data are the original predictors, and embedded values, such as the PCA components, are the derived predictors. 
```

feels like it drives the point home more. Especially since you rarely would have a "day of the week" predictor that wasn't created by a date variable anyways

chapters/feature-selection.qmd
```diff
@@ -0,0 +1,525 @@
...
+
+This idea will arise again in @sec-importance when we discuss _variable importance scores_ for explaining models. 
+
+## Overfitting the Feature Set {#sec-selection-overfitting}
+
+We’ve previously discussed overfitting, where a model finds patterns in the training data that are not reproducible. It is also possible to _overfit the predictor set_ by finding the training set predictors that appear to be connected to the outcome but do not show the same relationship in other data. 
+
+This is most likely to occur when the number of predictors ($p$) is much larger than the training set size ($n_{tr}$) or when no external validation is used to verify performance. @Ambroise2002p1493 shows reanalyses of high-dimensional biology experiments using a backward selection algorithm for removing predictors. In some cases, they could find a model and predictor subset with perfect accuracy even when the training set outcome data were randomly shuffled.
```
EmilHvitfeldt, line 147: no note, just very funny

chapters/feature-selection.qmd
```diff
@@ -0,0 +1,525 @@
...
+
+For example, if we use a resampling scheme, we have to repeat the feature selection process within each resample. If we used 10-fold cross-validation, each of the 10 analysis sets (90% of the training set) would have its own set of selected predictors. 
+
+@Ambroise2002p1493 demonstrated that when feature selection was enacted once, and then the model with this subset was resampled, they could produce models with excessively optimistic performance metrics. However, when the performance statistics were computed with multiple realizations of feature selection, there was very little false optimism. 
+
+Another improper example of data usage that can be found in the literature, as well as in real-life analyses, occurs when the entire data set is used to select predictors _then_ the initial split is used to create the training and test sets. This “bakes in” the false optimism from the start^[Recall out mantra that “Always have a separate piece of data that can **contradict what you believe**.”].
```
EmilHvitfeldt, line 155: > This “bakes in” the false optimism from the start

data leakage correct?

chapters/feature-selection.qmd
```diff
@@ -0,0 +1,525 @@
...
+
+@Ambroise2002p1493 demonstrated that when feature selection was enacted once, and then the model with this subset was resampled, they could produce models with excessively optimistic performance metrics. However, when the performance statistics were computed with multiple realizations of feature selection, there was very little false optimism. 
+
+Another improper example of data usage that can be found in the literature, as well as in real-life analyses, occurs when the entire data set is used to select predictors _then_ the initial split is used to create the training and test sets. This “bakes in” the false optimism from the start^[Recall out mantra that “Always have a separate piece of data that can **contradict what you believe**.”]. 
+
+However, performing feature selection in each resample can drastically increase the computational cost, especially if the model has many tuning parameters. For this reason, we tend to focus on methods that _atuomatically select features_ during the normal process of training the model (as discussed below).
```
EmilHvitfeldt, line 157: > can drastically increase the computational cost, especially if the model has many tuning parameters.

This is a multiplicative increase in computational cost correct? i get the sentiment but "drastically" feels like a exponential increase type of word.

chapters/feature-selection.qmd
```diff
@@ -0,0 +1,525 @@
...
+
+However, performing feature selection in each resample can drastically increase the computational cost, especially if the model has many tuning parameters. For this reason, we tend to focus on methods that _atuomatically select features_ during the normal process of training the model (as discussed below). 
+
+One scheme to understand if we are overfitting to the predictors is to create one or more artificial "sentinel predictors" that are random noise. Once the analysis that ranks/selects predictors is finished, we can see how high the algorithm ranks these deliberately irrelevant predictors. It is also possible to create a rule where predictors that rank higher than the sentinels are selected to be included in the model. 
+
+## Unsupervised Selection {#sec-unsupervised-selection}
+
+There are occasions where the values of the predictors are unfavorable or possibly detrimental to the model. The simplest example is a zero-variance column, where all of the rows of the training set have the same value. There is no information in these data and, although the model might be able to tolerate such a predictor, taking the time to identify and remove such columns is advisable.
```
EmilHvitfeldt, line 163:
```suggestion
There are occasions where the values of the predictors are unfavorable or possibly detrimental to the model. The simplest example is a zero-variance column, where all of the values of a predictor in the training set have the same value. There is no information in these predictors and, although the model might be able to tolerate such a predictor, taking the time to identify and remove such columns is advisable. 
```

chapters/feature-selection.qmd
```diff
@@ -0,0 +1,525 @@
...
+
+There are occasions where the values of the predictors are unfavorable or possibly detrimental to the model. The simplest example is a zero-variance column, where all of the rows of the training set have the same value. There is no information in these data and, although the model might be able to tolerate such a predictor, taking the time to identify and remove such columns is advisable. 
+
+But what about situations where only one or two rows have non-zero values? It is difficult to filter on variance values since the predictors could have different units or ranges. The underlying problem occurs when very few unique values exist (i.e., “coarse data”), and one or two predictor values capture most of the rows. Perhaps we can identify this situation. 
+
+Consider a hypothetical training set of 500 data points with a count-based predictor that has mostly zero counts. However, there are three data points with a count of 1 and one instance with a count of 2. The variance of these data is low, but it would be difficult to determine a cutpoint for removal. @kuhn2008building developed an algorithm to find **near-zero predictors**, as defined by two criteria:
```
EmilHvitfeldt, line 167:
```suggestion
Consider a hypothetical training set of 500 data points with a count-based predictor that has mostly zero counts. However, there are three data points with a count of 1 and one instance with a count of 2. The variance of these data is low, but it would be difficult to determine a cutpoint for removal. @kuhn2008building developed an algorithm to find **near-zero variance predictors**, as defined by two criteria:
```

chapters/feature-selection.qmd
```diff
@@ -0,0 +1,525 @@
...
+  revC = FALSE,
+  label_format_fun = function(...) round(..., digits = 2),
+  limits = c(-1, 1),
+  margins = c(50, 215, 10, 150)
+)
+```
+
+Similar redundancy can occur in categorical predictors. In this case, measures of similarity can be used to filter the predictors in the same way. 
+
+Unsupervised filters are only needed in specific circumstances. Some models (such as trees) are resistant to the distributions of the predictor’s data, while others are excessively sensitive.
```
EmilHvitfeldt, line 219: Maybe add a note that while zero-variance filtering is only needed sometimes, it never hurts (and is VERY fast)

chapters/feature-selection.qmd
```diff
@@ -0,0 +1,525 @@
...
+)
+```
+
+Similar redundancy can occur in categorical predictors. In this case, measures of similarity can be used to filter the predictors in the same way. 
+
+Unsupervised filters are only needed in specific circumstances. Some models (such as trees) are resistant to the distributions of the predictor’s data, while others are excessively sensitive. 
+
+## Classes of Supervised Methods {#sec-selection-methods}
+
+There are three main strategies for _supervised_ feature selection. The first and most effective was discussed: automatic selection occurs when a subset of predictors is determined during model training. We’ll discuss this more in the next section.
```
EmilHvitfeldt, line 223: nit-pick, we talk about three main strategies, but this one doesn't have a bolded word and the next two sections does

chapters/feature-selection.qmd
```diff
@@ -0,0 +1,525 @@
...
+
+::: {.important-box}
+This shows that we can often reduce the number of features for free. 
+:::
+
+Different models view the training data differently, and the list of relevant predictors will likely change from model to model. When we fit a MARS model (@sec-mars-reg) to the data, only `r nrow(mars_active)` original predictors were used. If we use automatic selection methods to understand what is important, we should fit a variety of different models and create a consensus estimate of which predictors are "active" for the data set.  
+
+## Wrapper Methods  {#sec-wrappers}
+
+Wrapper methods [@kohavi1998wrapper] use an overarching search method to optimize binary indicators, one for each potential predictor, that control which predictors are given to the model.
```
EmilHvitfeldt, line 265:
```suggestion
Wrapper methods [@kohavi1998wrapper] use an overarching search method to optimize **binary indicators**, one for each potential predictor, that control which predictors are given to the model.
```

Very first time i read this i thought for a moment we were talking dummy variables

chapters/feature-selection.qmd
```diff
@@ -0,0 +1,525 @@
...
+
+Two global search methods described in @sec-iterative, genetic algorithms and simulated annealing, can also be used. In this instance, the search space is the predictor space represented as binary indicators. The same process applies, though. For example, a GA would contain a population of different subsets, each producing a fitted model. The performance values enable the creation of the next generation via selection, reproduction, and mutation. In the case of simulated annealing, an initial subset is created and evaluated, and perturbations are created by randomly altering a small set of indicators (to create a "nearby" subset). Suboptimal subsets can be accepted similarly to our application for tuning parameters.
```
EmilHvitfeldt, line 269:
```suggestion
Two global search methods described in @sec-iterative, genetic algorithms (GA) and simulated annealing, can also be used. In this instance, the search space is the predictor space represented as binary indicators. The same process applies, though. For example, a GA would contain a population of different subsets, each producing a fitted model. The performance values enable the creation of the next generation via selection, reproduction, and mutation. In the case of simulated annealing, an initial subset is created and evaluated, and perturbations are created by randomly altering a small set of indicators (to create a "nearby" subset). Suboptimal subsets can be accepted similarly to our application for tuning parameters. 
```

nit-pick. You did it for recursive feature elimination (RFE). but i can also see why you don't since you properly already explained the abbreviation in @sec-iterative

chapters/feature-selection.qmd
```diff
@@ -0,0 +1,525 @@
...
+
+## Filter Methods {#sec-filters}
+
+Filters can use any scoring function that quantifies how effectively the predictor predicts the outcome [@kohavi1998wrapper;@duch2006filter]. One method is to use a univariate statistic to evaluate each predictor independently. For example, the area under the ROC curve (@sec-roc) can be computed for a binary classification outcome and a set of numeric predictors. The filter could be applied using these values by pre-defining a threshold for the statistic that defines "important enough." Alternatively, the top $p^*$ predictors can be retained and used in the supervised model.
```
EmilHvitfeldt, line 333: I have started to heavily lean towards suggestion that we use thresholds instead of counts. Due to the fact that it removes issues with ties. "asking for top 10 but top 9, 10, 11, 12 have same value"

chapters/feature-selection.qmd
```diff
@@ -0,0 +1,525 @@
...
+
+The example of using ROC curves raises an interesting question. Should we rank the predictors by their area under the curve or should we convert that analysis to a p-value to see if the area under the curve is greater than 0.5 [@hanley1983method]? Our advice is to do the former. Hypothesis tests and p-values have their uses but the AUC estimates themselves are probably more numerically stable. To factor in uncertainty, we could filter on the lower confidence bound of the AUC.  This is generally true for almost any ranking statistic that has a corresponding p-value. 
+
+Multivariate methods can also be used. These simultaneously compute measures of importance for all predictors. This is usually preferred since univariate statistics can be compromised when the predictors are highly correlated and/or have important interaction effects.
```
EmilHvitfeldt, line 337: Is there a computational argument to be used on whether we use multivariate methods or univariate?

chapters/feature-selection.qmd
```diff
@@ -0,0 +1,525 @@
...
+    x = NULL,
+    y = "# Selections in 10 Folds",
+    title = "(b) Selection Frequency"
+  ) +
+  scale_y_continuous(breaks = pretty_breaks())
+
+(p_rmse | p_terms) + plot_layout(widths = c(4, 3))
+```
+
+Recall that the importance scores are computed on different analysis sets. For the best model, @fig-mlp-rf(b) contains the frequency at which each predictor was selected. For the most part, the importance scores are pretty accurate; almost all informative features are consistently selected across folds. When the final model was fit on the entire training set, `r cli::format_inline("{num_noise} noise predictor{?s} {?was/were}")`  included in the model along with predictors `r cli::format_inline("{non_noise}")`. These results are unusually clean, mostly due to the large sample size and relatively small predictor set.
```
EmilHvitfeldt, line 515: I see now that "1 noise predictor was included in the model" was computer generated. I feel like it would need a sentence pointing out why.

it almost read like it was a active choice to include one, but it should really be "our model wasn't able to remove this one"

EmilHvitfeldt (commented):
Looks good! few notes

chapters/feature-selection.qmd
topepo: We used the acronym in a previous chapter but it's a good idea to do it again

chapters/feature-selection.qmd
topepo: Yep! 💯

chapters/feature-selection.qmd
topepo: It depends on the method. I think my preference is based on efficacy and not efficiency.

PR #76, opened 2025-04-07
Added Comparing Models chapter

kjell-stattenacity:


topepo:
I made some preliminary, general updates to the code. 

I'd like to make two changes to the section on Resampled Data Sets. First, let's change the example (just for this section) to use the forestation data from upcoming chapters (other data sets would also work for me). My thinking is:

- Originally, we used a validation set (and one different model). I want to be consistent with the original analysis of the data. 
- The tiny difference in models (in units of seconds) doesn't effectively highlight the ROPE analysis. 

Second, I'd like the hierarchical model to be the same random intercept model from the [racing section](https://aml4td.org/chapters/grid-search.html#sec-racing) (aka equation 11.1). This model can be used for both Frequentist and Bayesian models, and we can still make Tukey (and other) p-value corrections. The blocking model is not bad, but all of our other analyses (and code) use the random intercept model. Let me know if this is an issue with the TOST test.

PR #87, opened 2025-05-24
use census data to get county for forested data

topepo:
This will be helpful for dealing with things like categorical splits in trees etc. 

We lose 11 data points (out of thousands)

image


topepo:
@kjell-stattenacity: the new object for `last_fit()` is called `forested_split` and is in `RData/forested_data.RData`

PR #91, opened 2025-06-17
Comparing models chapter

topepo:


chapters/comparing-models.qmd
```diff
@@ -938,75 +996,67 @@ To adapt this idea to a single validation set, we treat the validation set as ou
...
-  ) +
-  ylab("MAE")
+glmnet_validation_res <-
+    glmnet_final_wflow %>%
+    last_fit(class ~ ., split = forested_split, metrics = class_metrics)
+
+glmnet_validation_boot <- int_pctl(glmnet_validation_res, alpha = 0.10)
+
+
+c50_validation_res <-
```
topepo, line 1026: I'd like to make a function that also does the traditional CI for a proportion and show how the bootstrap mirrors that. This will give them some confidence that it actually works

PR #94, opened 2025-07-19
Forestation data introduction

topepo:
Some initial analysis that will help facilitate feature engineering


topepo:
@simonpcouch There are a lot of ancillary changed files since this PR branch is a trimmed version of additional content. You can focus on the changes in `cls-linear.qmd`.

chapters/cls-linear.qmd
```diff
@@ -11,41 +11,259 @@ knitr:
...
+
+Individually, how does each of these features appear to relate to the outcome? To assess this, we binned each predictor into roughly 20 groups based on percentiles and used these groups (each containing about `r floor(mean(percentiles$total))` locations) to compute the rate of forestation and 90% confidence intervals^[Binning is used here as a visualization tool; we re-emphasize that converting numeric predictors into categorical features is problematic.]. As an exploratory tool, we can use these binned versions of the data to see potential relationships with the outcome. @fig-forest-percentiles shows the profiles. Note that there is enough data in each bin to make the confidence intervals very close to the estimated rates.
```
simonpcouch, line 175: re:

>  Note that there is enough data in each bin to make the confidence intervals very close to the estimated rates. 

It's surprising to me that the intervals are so tight around `year` even when there's so much variation in the mean in the upper 25%?


Screenshot 2025-07-21 at 3 01 41 PM

Seeing that binned variation, I would have guessed that the intervals would be wider.

simonpcouch (approved):
This is a kickass section. I dig it.

Can't remember what sorts of introductions to the data already exist in this book, but I do think a quick viz of the plot sampling design would make it clearer why we're not so worried about autocorrelation here. [This page](https://www.srs.fs.usda.gov/pubs/gtr/gtr_srs080/gtr_srs080.pdf#page=31) in the Green Book has a nice example.

chapters/cls-linear.qmd
```diff
@@ -11,41 +11,259 @@ knitr:
...
+    county = factor(county),
+    county = reorder(county, rate),
+    `# Locations` = total
+  )
+```
+
+As with any ML project, we conduct preliminary exploratory data analysis to determine whether any data characteristics might affect how we model them. @tbl-forested-numeric has statistical and visual summaries of the `r ncol(forested_train) - 2` numeric predictors using the training data. Several of the predictors exhibit pronounced skew (right or left leaning). By coercing the distributions of some predictors to be more symmetric, we might gain robustness and perhaps an incremental improvement in performance (for some models). 
+
+Also, the annual minimum temperature, dew temperature, January minimum temperature, and maximum vapor show bimodality in the distributions. The year of inspection is also interesting; the data collection was sparse before 2011, and subsequent years contain a few hundred data points per year before beginning to drop off in 2021. These characteristics are not indicative of problems with data quality, but it can be important to know that they exist when debugging why a model is underperforming or showing odd results.
```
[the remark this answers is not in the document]
topepo: I assumed it was like how we spend time on packages: blocks of time where there is solid focus in between times of silence...

chapters/cls-linear.qmd
topepo: I felt the same way and initially worried that I used `conf.level = 0.1` instead of `conf.level = 0.9` but it is correct. I tend to use 90% intervals (more narrow than 95%) so that is part of it. 

There are about 230 rows per point so the +/- is about 4%

PR #102, opened 2025-09-15
logistic regression part 1

topepo:
Adds

 - basic logistic regression
 - GLM introduction
 - fit on forestation data
 - diagnostics
 - interpretation

Next PR is for multicollinearity penalized models, and Bayes


chapters/cls-linear.qmd
```diff
@@ -236,32 +330,779 @@ forested_hstats_text |>
...
+  mutate(
+    indicator = ifelse(class == "Yes", 1, 0),
+    error = indicator - .pred_Yes,
+    dev_1 = indicator * log(.pred_Yes),
+    dev_2 = (1 - indicator) * log(1 - .pred_Yes),
+    deviance = sign(error) * sqrt((-2 * (dev_1 + dev_2)))
+  ) |>
+  inner_join(forested_train |> add_rowindex() |> select(-class), by = ".row")
+
+logistic_resdiduals |>
```
topepo, line 868: Maybe I should overlay the residuals on a map?

chapters/cls-linear.qmd
```diff
@@ -236,32 +330,805 @@ forested_hstats_text |>
...
 
 The next section will describe a mainstay of machine learning models for two classes: logistic regression. 
 
 ## Logistic Regression {#sec-logistic-reg}
 
+Logistic regression is a classification model that can be used when there are $C = 2$ classes. As with all binary classification models, we want to accurately estimate the probability of an event^[We'll assume that class $j = 1$ is the class that is considered the event, for $j = 1, \ldots, C$ classes.]. As discussed in the previous chapter, the probability parameter $\pi$ should be between zero and one, and we often frame the problem via the Bernoulli distribution^[We often talk about logistic regression as being driven by a **Binomial** distribution. Either distribution can be used for this model, but the Binomial is used for _grouped_ data where each row of the data set has $n_i$ "trials" and $r_i$ "successes". In ML, we are not usually in this situation and have a single outcome result per row, hence why we use the Bernoulli (where $n_i = 1$).] (i.e., $y \sim Bernoulli(\pi)$). The likelihood function is:
```
kjell-stattenacity:
```suggestion
Logistic regression is a classification model that can be used when there are $C = 2$ classes. Like all binary classification models, its goal is to accurately estimate the probability of an event occurring^[We'll assume that class $j = 1$ is the class that is considered the event, for $j = 1, \ldots, C$ classes.]. As discussed in the previous chapter, the probability parameter $\pi$ must be between zero and one, and we often frame the problem via the Bernoulli distribution^[We often talk about logistic regression as being driven by a **Binomial** distribution. Either distribution can be used for this model, but the Binomial is used for _grouped_ data where each row of the data set has $n_i$ "trials" and $r_i$ "successes". In ML, we are not usually in this situation and have a single outcome result per row, hence why we use the Bernoulli (where $n_i = 1$).] (i.e., $y \sim Bernoulli(\pi)$). The likelihood function is:  
```

chapters/cls-linear.qmd
```diff
@@ -236,32 +330,805 @@ forested_hstats_text |>
...
+
+Overall, for logistic regression, maximum likelihood estimation generates a relatively stable estimator. While it can be sensitive to outliers and high influence points [@pregibon1981logistic], a limited amount of data jittering or omissions will not widely affect the parameter estimates unless the training set is fairly small. We’ll see an exception below in @sec-logistic-multicollinearity,  where specific characteristics of the _training data_ make the estimators unstable.   
+
+Statistically, the MLEs are estimates that best explain the data. They are also unbiased estimates of the true parameters^[Recall from @sec-variance-bias that unbiased estimates might have high variance. For the forestation data, we'll see that some of our logistic regression parameter estimates have irrationally large variances. Later in this chapter, biased methods (via regularization) are also discussed.].
```
kjell-stattenacity:
```suggestion
From a statistical perspective, MLEs yield parameter estimates that maximize the probability of observing the given data under the logistic model. They are also unbiased estimates of the true parameters.  However, they may suffer from high variance.^[Recall from @sec-variance-bias that unbiased estimates might have high variance. For the forestation data, we'll see that some of our logistic regression parameter estimates have irrationally large variances. Later in this chapter, we will introduce biased estimation methods (via regularization) to reduce the variance and improve predictive performance.].
```

chapters/cls-linear.qmd
```diff
@@ -236,32 +330,805 @@ forested_hstats_text |>
...
+
+::: {.important-box}
+Before proceeding further, let's talk about the linear predictor. The "linear" in "linear predictor" and "generalized linear model" means that the model is _linear in the parameters_ $\boldsymbol{\beta}$. This does **not** mean that the model can only produce linear class boundaries. Our model terms $x_{ij}$ can represent any function of one or more parameters. We can use the feature engineering tools shown in Chapters [-@sec-numeric-predictors] through [-@sec-interactions-nonlinear] to create a better model. 
+:::
+
+For example, in @fig-longitude, we saw that using the longitude values as our single predictor $x$ didn't work well. When plotted on the logit scale, there is a clear nonlinear function in the binned rates. We know how to estimate a general nonlinear function of a feature: spline functions. @fig-longitude can also show the results when ten natural spline features are used in the model. Visually, we can see that this expansion is far more effective at predicting the outcome.
```
kjell-stattenacity:
```suggestion
For example, in @fig-longitude, we saw that using the longitude values as our single predictor $x$ didn't work well. When plotted on the logit scale, there is a clear nonlinear function in the binned rates.  To model this more flexibly, we incorporate spline-based features. Specifically, @fig-longitude displays the results when ten natural spline features are used in the model. Visually, we can see that this expansion improves the model's accuracy at predicting the outcome.
```

chapters/cls-linear.qmd
```diff
@@ -236,32 +330,805 @@ forested_hstats_text |>
...
+  extract_workflow_set_result(glm_best$wflow_id) |>
+  int_pctl(times = 2000, metrics = cls_mtr)
+```
+
 #### Forestation Model Development {.unnumbered}
 
-#### Interpretation {.unnumbered}
+We can assess each of these feature engineering steps by fitting logistic regression pipelines that sequentially add steps. Our five models are:
+
+ 1. A **simple** preprocessor that only converts the `r length(levels(forested_train$county))` counties to binary indicators and removes potential zero-variance predictors.
```
kjell-stattenacity:
```suggestion
 1. **Simple model**: Convert each of the `r length(levels(forested_train$county))` counties to binary indicators and drop any predictors with zero-variance. 
```

chapters/cls-linear.qmd
```diff
@@ -236,32 +330,805 @@ forested_hstats_text |>
...
+```
+
 #### Forestation Model Development {.unnumbered}
 
-#### Interpretation {.unnumbered}
+We can assess each of these feature engineering steps by fitting logistic regression pipelines that sequentially add steps. Our five models are:
+
+ 1. A **simple** preprocessor that only converts the `r length(levels(forested_train$county))` counties to binary indicators and removes potential zero-variance predictors. 
+ 2. We can add a **normalization** step that transforms the numeric predictors using the ORD transformation. 
+ 3. We'll replace the county code dummy variable operation in model #2 with an effect **encoding** step.
```
kjell-stattenacity:
```suggestion
 3. **Encoding model**:  Build on the normalization model by replacing the county dummy indicators with effect encoding.
```

chapters/cls-linear.qmd
```diff
@@ -236,32 +330,805 @@ forested_hstats_text |>
...
+#| include: false
+logistic_mtr <-
+  forest_res |>
+  extract_workflow_set_result(
+    "encoded + transformations + splines_logistic_mle"
+  ) |>
+  collect_metrics(summarize = TRUE)
+```
+
+Along with the Brier score values of `r signif(logistic_mtr$mean[logistic_mtr$.metric == "brier_class"], 3)`, the spline pipeline (i.e., Model #5) had the following resampling estimates: ROC AUC = `r signif(logistic_mtr$mean[logistic_mtr$.metric == "roc_auc"], 3)`, a PR AUC = `r signif(logistic_mtr$mean[logistic_mtr$.metric == "pr_auc"], 3)`, and a cross-entropy = `r signif(logistic_mtr$mean[logistic_mtr$.metric == "mn_log_loss"], 3)`. During resampling, there were `r logistic_mtr$n[1]` resamples, leading to `r logistic_mtr$n[1]` ROC curves, and so on. We can pool the `r logistic_mtr$n[1]` sets of assessment set predictions and produce _approximate_ ROC and calibration curves. These are shown in @fig-forest-logistic-diag. The calibration curve indicates that the model reliably predicts samples with accuracy  The ROC curve shows good separation between classes but is otherise unremarkable.
```
kjell-stattenacity:
```suggestion
Along with the Brier score values of `r signif(logistic_mtr$mean[logistic_mtr$.metric == "brier_class"], 3)`, the spline pipeline (i.e., Model #5) had the following resampling estimates: ROC AUC = `r signif(logistic_mtr$mean[logistic_mtr$.metric == "roc_auc"], 3)`, a PR AUC = `r signif(logistic_mtr$mean[logistic_mtr$.metric == "pr_auc"], 3)`, and a cross-entropy = `r signif(logistic_mtr$mean[logistic_mtr$.metric == "mn_log_loss"], 3)`. During resampling, there were `r logistic_mtr$n[1]` resamples, leading to `r logistic_mtr$n[1]` ROC curves and corresponding performance metrics. By pooling predictions from all `r logistic_mtr$n[1]` assessment sets, we constructed _approximate_ ROC and calibration curves (@fig-forest-logistic-diag). The calibration curve indicates that the model reliably predicts samples with accuracy  The ROC curve shows good class separation, although it's performance is otherwise unremarkable. 
```

chapters/cls-linear.qmd
```diff
@@ -236,32 +330,805 @@ forested_hstats_text |>
...
+  coord_obs_pred()
+
+logitic_cal + logitic_roc
+```
+
+From here, we should investigate which terms had the greatest influence on the model. Were the 10 interactions that we identified important or statistically significant? Should we have added more or less? We might plot the predicted values to see if there are specific geographic locations that corresponded to the poorly calibrated values. Perhaps we could also refine the model by using a supervised feature filter, and so on. 
+
+### Examining the Model {#sec-logistic-diagnostic}
+
+There are many ways to appraise the model. One (unfortunate) focus is often on the statistical significance of the model terms. If the model’s purpose is solely for interpretation, this is understandable. For ML applications, p-values are not particularly insightful since they don’t necessarily correlate with predictive performance. This is especially true for our logistic model, which contains numerous spline terms; knowing whether the fourth spline term has a small p-value probably isn’t insightful. However, we’ll return to the model inferential capacity shortly as a diagnostic tool.
```
kjell-stattenacity:
```suggestion
There are many ways to evaluate a model’s performance. One common, but often misplaced, focus is on the statistical significance (p-values) of individual model terms. If the model’s primary goal is interpretation, this may make sense. However, for ML applications, p-values are usually not very informative, because they often do not correlate with how well the model will make predictions. This is especially true in our logistic regression model, which includes many spline basis functions: knowing that the fourth spline term has a small p-value tells us little about overall predictive power. That said, we will revisit inference later as a diagnostic tool.
```

chapters/cls-linear.qmd
```diff
@@ -236,32 +330,805 @@ forested_hstats_text |>
...
+
+Our suggestion is to assess the model’s quality by using the held-out predictions to conduct *residual analysis* by finding systematic patterns in those values. The most basic residual is the simple difference in the "observed" and predicted probabilities: $e_{ic} = y_{ic} - \widehat{p}_{ic}$. The issue with these values is that the variance of the probability estimates changes over their [0,1] range. A residual of 0.1 has a different context when $\widehat{p}_i = 0.01$ compared to $\widehat{p}_i = 0.5$.  To normalize these values, we can use _Pearson residuals_, which are $e_{ic}$  divided by the standard deviation:
```
kjell-stattenacity:
```suggestion
Instead, for assessing model quality, we recommend using the held-out predictions to conduct *residual analysis*, looking for systematic deviations in prediction errors. The most basic residual is the simple difference between the "observed" outcome and the predicted probability: $e_{ic} = y_{ic} - \widehat{p}_{ic}$.  However, these residuals have heteroscedastic variance--the probability estimates changes over their [0,1] range. A residual of 0.1 has a different context when $\widehat{p}_i = 0.01$ compared to $\widehat{p}_i = 0.5$.  To normalize these values, we can use _Pearson residuals_, which are $e_{ic}$  divided by the standard deviation: 
```

chapters/cls-linear.qmd
```diff
@@ -236,32 +330,805 @@ forested_hstats_text |>
...
+
+Alternatively, we can look at _deviance residuals_. The deviance is a likelihood-based comparison between two models: our current model and one that is _saturated_ where the model perfectly fits the training data. For logistic regression, these take the form: 
+
+$$
+e^d_{ic} = \text{sign}(e_{ic}) \left[-2(y _{ic}\, \text{log}(\widehat{p} _{ic}) + (1 - y _{ic})\, \text{log}(1 - \widehat{p}_{ic}))\right]^{1/2}
+$$
+
+See @pregibon1981logistic and  @agresti2015foundations for discussions.  
+
+In our application, it would be relevant to plot residuals versus values of individual predictors. If we see systematic, non-random patterns, it might indicate that how we represent that predictor in the model is inadequate. For example, in @fig-longitude-resdiauls the deviance residuals are plotted against longitude. These residuals are computed with the holdout predictions for each resampling fold.
```
kjell-stattenacity:
```suggestion
In practice, we should plot residuals against each predictor’s values to check for systematic patterns. If the residuals show non-random structure (e.g. curved or clustered patterns), it suggests that the predictor may not be represented properly in the model. For example, in Figure @fig-longitude-residuals we plot deviance residuals versus longitude, using predictions from each resampling fold on the held-out data.
```

chapters/cls-linear.qmd
```diff
@@ -236,32 +330,805 @@ forested_hstats_text |>
...
+  geom_point(alpha = 1 / 2, cex = 1) +
+  facet_wrap(~Model) +
+  scale_color_manual(values = c("#218239", "#d4ad42")) +
+  labs(x = "Longitude", y = "Deviance Residiual") +
+  theme(legend.position = "top")
+```
+
+In contrast, the right-hand panel has many more residuals near zero and significantly attenuated spikes/patterns. This is indicative of an improvement of model fit. However, some points have absolute residuals _larger_ than the simpler model. Overall, though, the model with spline terms has fewer systematic patterns. 
+
+Given the spatial nature of these data, mapping the residuals can be instructive. It could show us locations that tend to be poorly predicted. We can also determine the largest (absolute) residuals and investigate what might be different about these particular data points.
```
kjell-stattenacity:
```suggestion
Given the spatial structure of our data, mapping residuals geographically is also useful—it can reveal particular locations where predictions are consistently poor. We can identify those points with the largest absolute residuals and investigate what makes them different. 
```

chapters/cls-linear.qmd
```diff
@@ -236,32 +330,805 @@ forested_hstats_text |>
...
+
+set.seed(1805)
+pdp_longitude <- model_profile(
+  explainer_logistic,
+  N = 1000,
+  variables = "longitude"
+)
+```
+
+Methods are also available to help explain a model or a specific prediction, independent of the type of model used. The field of **explainable machine learning** uses tools related to the H-statistic previously described in @sec-interactions-detection. These will be discussed in detail in @sec-explanations, but we’ll preview the technique here.
```
kjell-stattenacity:
```suggestion
In addition to fitting models, there are methods to explain how a model or a specific prediction, works, regardless of the model type.  These tools are part of the field of **explainable machine learning**.  Some, like the H-statistic for detecting interactions (see @sec-interactions-detection) help quantify effects of predictors. These will be discussed in detail in @sec-explanations, but here is a preview. 
```

chapters/cls-linear.qmd
```diff
@@ -236,32 +330,779 @@ forested_hstats_text |>
...
 
 ### Bayesian Estimation {#sec-logistic-bayes} 
 
-## Generalized Additive Models {#sec-cls-gam}
-
 ## Multinomial Regression {#sec-multinomial-reg}
 
+## Generalized Additive Models {#sec-cls-gam}
+
 ## Discriminants {#sec-lda}
```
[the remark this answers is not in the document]
topepo: Maybe not. The two neural sections in the next chapter are going to take up a lot of space so let's keep the GAM parts here.

PR #103, opened 2025-09-30
Multilayer Perceptrons

topepo:
The first of two chapters on neural networks. 

PDF for anyone who doesn't want to compile the books: [Neural Networks via Multilayer Perceptrons.pdf](https://github.com/user-attachments/files/22624020/Neural.Networks.via.Multilayer.Perceptrons.pdf)

### DIFF
--- a/chapters/cls-nonlinear.qmd
+++ b/chapters/cls-nonlinear.qmd
@@ -43,6 +43,14 @@ load("../RData/optimizers.RData")
 source("../R/setup_2D_GD.R")
 ```
 
+This chapter focuses on several classes of models that naturally produce nonlinear boundaries, such as neural networks, K-nearest neighbors, and support vector machines. A separate class of models, which use tree structures, also has inherently nonlinear classification boundaries. We'll discuss those models in @sec-cls-trees and @sec-cls-ensembles. 
+
+In this chapter, several recurring motifs are present across different types of models. First, this involves the use of an operation that utilizes the positive part of a function, known as "hinge" or "rectifier" functions. We'll see these in @sec-fda and @sec-nnet-activation. Another element that will be encountered across different models is the kernel function, previously seen in @sec-gp. These functions are used to quantify the distance or similarity between data points (or, more specifically, vectors). Both the K-nearest neighbor and the support vector machine (SVM) models rely on these functions. Finally, the dot product between vectors plays a critical role in the important _attention_ mechanism in neural networks (@sec-attention) as well as in SVM models (@sec-cls-svm). 
+
+This chapter contains _two_ sections on neural networks. The first describes the most basic version of these models, called the _multilayer perceptron_. The fundamentals of these models are discussed in this section. The second section explores various methodologies to enhance the performance of neural network models for tabular data, similar to their effectiveness in predicting images and text. At the time of this writing, substantial research on this topic is underway in the field. This section will also discuss one of the most effective tools for most deep learning models: attention-based transformers.
+
+To get started, the first chapter summarizes an old, traditional model for classification, linear discriminant analysis (LDA), and shows how it can be effectively modified to produce nonlinear class boundaries.   
+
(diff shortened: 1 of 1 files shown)


chapters/cls-nonlinear.qmd
```diff
@@ -41,11 +53,885 @@ load("../RData/forested_data.RData")
...
+
+Note that the learning rate $\alpha$ now has a subscript. They also recommend trying a cyclic learning rate scheduler with AdamW.  
+
+#### Other Methods {.unnumbered} 
+
+As previously mentioned, there are numerous other variations of SGD that will not be discussed here, including AdaMax [@kinga2015method], Nadam [@dozat2016], MadGrad [@defazio2021], and others.  
+
+### Newton's Method {#sec-newtons-method}
+
+As previously shown, Newton’s method (a.k.a., the Newton–Raphson method) uses both gradients and the Hessian matrix to update parameters. This approach can be very effective when the local loss surface can be approximated by a second-degree polynomial. When the local search includes a stationary point, the signs of the eigenvalues of the Hessian can characterize it: all non-negative imply a maximum, all non-positive imply a minimum, and mixed signs indicate a saddle point.
```
topepo, line 739: Originally, I was getting the best results from L-BFGS. However, I eventually figured out that `torch::optim_lbfgs()$step()` does _all_ of the steps instead of stopping at the end of an iteration. It turns out that, for `I` iterations, I was running `I` complete L-BFGS runs, each with 20 iterations. 

Not sure what to do about this in brulee just yet.

chapters/cls-nonlinear.qmd
```diff
@@ -41,11 +53,885 @@ load("../RData/forested_data.RData")
...
+
+These connections between units on different layers continue as layers are added. Eventually, the last layer is the _output layer_. For classification, this usually has as many units as there are class levels in the outcome. @fig-nnet-cls shows a general architectural diagram for a basic model with a single hidden layer, where the lines indicate that a model parameter exists to connect the nodes..
```
dfalbel: nit

```suggestion
These connections between units on different layers continue as layers are added. Eventually, the last layer is the _output layer_. For classification, this usually has as many units as there are class levels in the outcome. @fig-nnet-cls shows a general architectural diagram for a basic model with a single hidden layer, where the lines indicate that a model parameter exists to connect the nodes. 
```

chapters/cls-nonlinear.qmd
```diff
@@ -41,11 +53,885 @@ load("../RData/forested_data.RData")
...
+
+#### Other Methods {.unnumbered} 
+
+As previously mentioned, there are numerous other variations of SGD that will not be discussed here, including AdaMax [@kinga2015method], Nadam [@dozat2016], MadGrad [@defazio2021], and others.  
+
+### Newton's Method {#sec-newtons-method}
+
+As previously shown, Newton’s method (a.k.a., the Newton–Raphson method) uses both gradients and the Hessian matrix to update parameters. This approach can be very effective when the local loss surface can be approximated by a second-degree polynomial. When the local search includes a stationary point, the signs of the eigenvalues of the Hessian can characterize it: all non-negative imply a maximum, all non-positive imply a minimum, and mixed signs indicate a saddle point. 
+
+One issue with the method is related to the Hessian. If there are $p$ parameters, the matrix needs to save $p(p+1)2$ parameters. For neural networks, $p$ can be in the millions, so the basic application of Newton’s method can be slow and might require infeasible amounts of memory (depending on the model size). We also need to invert the Hessian, making it even more costly.
```
dfalbel: Maybe missing the `/`?

```suggestion
One issue with the method is related to the Hessian. If there are $p$ parameters, the matrix needs to save $p(p+1)/2$ parameters. For neural networks, $p$ can be in the millions, so the basic application of Newton’s method can be slow and might require infeasible amounts of memory (depending on the model size). We also need to invert the Hessian, making it even more costly.
```

topepo: Yep

PR #119, opened 2026-05-23
BatchNorm

topepo:
A short section in within-model normalization before leading into residual networks.


chapters/cls-neural-nets.qmd
```diff
@@ -604,7 +604,7 @@ For non-numeric data, the tools described in @sec-categorical-predictors can be
...
+For numeric predictors, we suggest avoiding irrelevant predictors and highly collinear feature sets^[Although the need for preprocessing features prior to training a neural network has long been recognized, @lecun2002efficient and @wiesler2014mean re-emphasized that standardization (i.e., centering and scaling) and decorrelating are critical for the optimization algorithm to converge within a reasonable time. ]. @fig-irrelevant-predictors in @sec-feature-selection demonstrated the harm of including predictors unrelated to the outcome in a neural network. Multicollinearity can also affect these models, even if we do not invert the feature matrix or use the Hessian during optimization. Highly redundant predictors can result in the loss function having difficult characteristics and a large number of unnecessary parameters. Using an L<sub>2</sub> penalty or a feature extraction method can be very helpful.
```
kjell-stattenacity:
```suggestion
For numeric predictors, we suggest avoiding irrelevant predictors and highly collinear feature sets^[Although the need for preprocessing features prior to training a neural network has long been recognized, @lecun2002efficient and @wiesler2014mean re-emphasized that standardization (i.e., centering and scaling) and decorrelating are critical for the optimization algorithm to converge within a reasonable time. ]. @fig-irrelevant-predictors in @sec-feature-selection demonstrated the harm of including predictors unrelated to the outcome in a neural network. Multicollinearity can also affect these models, even if we do not invert the feature matrix or use the Hessian during optimization. Highly redundant predictors can result in the loss function having difficult characteristics, such as ill-conditioned curvature and flat or degenerate regions, and a large number of unnecessary parameters. Using an L<sub>2</sub> penalty or a feature extraction method can be very helpful. 
```

PR #123, opened 2026-07-27
General attention discussion

topepo:
More PRs to come for column and row attention. NOTE: all changes are in `chapters/cls-neural-nets.qmd`. [PDF for convenience](https://github.com/user-attachments/files/30425892/Neural.Network.Classifiers.attention.pdf)


chapters/cls-neural-nets.qmd
```diff
@@ -1143,20 +1180,19 @@ knn_roc <-
 
 ## Modern Tabular Network Models {#sec-cls-tab-net}
 
-The previous section describes classical neural networks for tabular data. However, there has been an increasing research focus on adapting modern components of networks designed for image or text data to improve tabular data models. 
-  
+The previous section describes classical neural networks for tabular data. However, there has been increasing research focus on adapting modern network components designed for image or text data to improve tabular data models. 
+
 A fair amount of research has been motivated by the observation that tree ensembles often outperform tabular neural network models and are generally lower maintenance. As such, some methodologies try to add direct analogs to tree operations. In other cases, tabular models incorporate the same ideas that have catapulted neural networks to greater effectiveness in generative AI: attention and in-context learning. This section will describe a handful of potential upgrades to neural networks to improve their performance on tabular data.
```
kjell-stattenacity:
```suggestion
A fair amount of research has been motivated by the empirical observation that tree ensembles often outperform tabular neural network models and often do so with less tuning and computational effort.  This observation has motivated two general strategies for improving tabular neural network models. The first embeds explicit analogs of tree operations within the network architecture. The second adapts to the tabular domain the innovations that have driven the success of neural networks in generative AI, specifically attention mechanisms and in-context learning. This section will examine potential upgrades to neural networks to improve their performance on tabular data.
```

chapters/cls-neural-nets.qmd
```diff
@@ -1204,25 +1239,172 @@ $$
...
- 
+
+Two other in-network normalizations to mention are LayerNorm [@ba2016layer] and RMSNorm [@zhang2019root]. Both of these normalize across the embeddings per sample. For example, LayerNorm computes a mean and standard deviation across features for a single row. RMSNorm also operates across features, but divides them by the square root of the sum of the squared values. Note that if the features have a distribution centered around zero, this is the same as the spatial-sign transformation [@eq-spatial-sign] that projects the data to a hypersphere of radius $\sqrt{p}$ (instead of 1.0). As such, RMSNorm imparts some robustness against unusually large feature values and constrains the range of the normalized values. 
+
 ### Residual Networks {#sec-cls-resnet}
 
+Residual networks, a.k.a. ResNet, modify the basic MLP architecture by adding skip connections: the input to each block is added directly to the block's output before passing to the next layer^[Except for the final layer.]. 
+
+Let's illustrate this using an MLP where the number of hidden units in each of the three layers is the same as the number of predictors ($p$), as shown on the left:
```
kjell-stattenacity:
```suggestion
To illustrate, let's consider an MLP in which each of the three hidden layers has the same number of units as there are predictors ($p$), as shown on the left-hand side of @eq-resid-net:
```

chapters/cls-neural-nets.qmd
```diff
@@ -1204,25 +1239,172 @@ $$
...
+\overbrace{
+\begin{alignedat}{3}
+\boldsymbol{H}_1 &{}={} & f_1(\boldsymbol{x}) &{}+{} &\boldsymbol{x} \notag \\
+\boldsymbol{H}_2 &{}={} & f_2(\boldsymbol{H}_1) &{}+{} &\boldsymbol{H}_1 \notag \\
+\boldsymbol{H}_3 &{}={} & f_3(\boldsymbol{H}_2) &{}+{} &\boldsymbol{H}_2 \notag \\
+\boldsymbol{Y} &{}={} & f_4(\boldsymbol{H}_3) \notag
+\end{alignedat}}^{\text{Residual Connections}}
+$$  {#eq-resid-net}
+
+For the MLP architecture, the idea is that ordinary feed-forward networks are tasked with learning, at each layer, a better representation of the outcome via $f_i(\cdot)$. This parameterization may be difficult to optimize because changes between layers may be very small, leading to potential issues with gradient magnitudes.
```
kjell-stattenacity:
```suggestion
For the MLP architecture, feed-forward network is tasked with learning, via $f_i(\cdot)$, a better representation of the data for predicting the outcome. This parameterization can be difficult to optimize because the changes from one layer to the next may be very small, which can lead to problems with the gradient magnitudes. 
```

kjell-stattenacity: Is this rewording accurate?

chapters/cls-neural-nets.qmd
```diff
@@ -1204,25 +1239,172 @@ $$
...
+\boldsymbol{H}_1 &{}={} & f_1(\boldsymbol{x}) &{}+{} &\boldsymbol{x} \notag \\
+\boldsymbol{H}_2 &{}={} & f_2(\boldsymbol{H}_1) &{}+{} &\boldsymbol{H}_1 \notag \\
+\boldsymbol{H}_3 &{}={} & f_3(\boldsymbol{H}_2) &{}+{} &\boldsymbol{H}_2 \notag \\
+\boldsymbol{Y} &{}={} & f_4(\boldsymbol{H}_3) \notag
+\end{alignedat}}^{\text{Residual Connections}}
+$$  {#eq-resid-net}
+
+For the MLP architecture, the idea is that ordinary feed-forward networks are tasked with learning, at each layer, a better representation of the outcome via $f_i(\cdot)$. This parameterization may be difficult to optimize because changes between layers may be very small, leading to potential issues with gradient magnitudes.  
+
+A residual network, shown on the right-hand side of @eq-resid-net, takes the current features prior to the hidden layer and adds them back into the embeddings (post-activation). In other words, the skip comes *before* $f_i(\cdot)$ and is added to the result *after* $f_i(\cdot)$ applies its activation. We'll use the term _residual block_ to describe the operations shown in each of the top three rows of the right-hand side.
```
kjell-stattenacity:
```suggestion
A residual network, shown on the right-hand side of @eq-resid-net, takes the features entering each hidden layer and adds them back into the embeddings (post-activation). In other words, the skip originates *before* $f_i(\cdot)$ and is added to the result *after* $f_i(\cdot)$ applies its activation. We will use the term _residual block_ to describe the operations shown in each of the top three rows of the right-hand side.
```

chapters/cls-neural-nets.qmd
```diff
@@ -1204,25 +1239,172 @@ $$
...
+
+ By adding the previous embeddings to this function, we've reframed the problem so that we are really learning $f_i(\boldsymbol{H}_{i-1}) = \boldsymbol{H}_i - \boldsymbol{H}_{i-1}$; a residual.
```
kjell-stattenacity:
```suggestion
 By adding the previous embeddings to  the block's output reframes the learning problem: rather than specifying the next representation, each block only needs to estimate how much the new representation should differ from the previous one, $f_i(\boldsymbol{H}_{i-1}) = \boldsymbol{H}_i - \boldsymbol{H}_{i-1}$, the _residual_ which gives this approach its name.
```

chapters/cls-neural-nets.qmd
```diff
@@ -1204,25 +1239,172 @@ $$
...
+
+ By adding the previous embeddings to this function, we've reframed the problem so that we are really learning $f_i(\boldsymbol{H}_{i-1}) = \boldsymbol{H}_i - \boldsymbol{H}_{i-1}$; a residual. 
+
+This process makes training easier, especially for models with many layers. Early practitioners assumed that performance could be increased simply by adding more layers. This was not the case. @he2016deep showed that large models paradoxically saw both training and test error increase as the number of layers increased. This isn't believed to be related to overfitting, but due to how gradients flow through the system during optimization. Residual connections address this by giving gradients a direct path back through the network: for layer $i$, the gradient is $f_i'(\boldsymbol{H}_{i-1}) + I$. If we add too much capacity to the model (e.g., more layers), it can simply estimate zero for the residual. Also, the extra term $I$ in the gradient prevents them from becoming too small across the network, even if $f_i'(\boldsymbol{H}_{i-1})$ has elements that are approximately zero. We'll describe other potential benefits to this model towards the end of this section.
```
kjell-stattenacity:
```suggestion
This reparameterization makes training easier, especially for models with many layers. Early practitioners assumed that performance could be increased simply by adding more layers, but this turned out not to be the case. @he2016deep showed that, paradoxically, large models saw both training and test error increase as the number of layers increased. This degradation isn't believed to be related to overfitting, but from the way gradients flow through the system during optimization.  Residual connections address this by giving gradients a direct path back through the network: for layer $i$, the gradient is $f_i'(\boldsymbol{H}_{i-1}) + I$. If we add too much capacity to the model (e.g., more layers), it can simply estimate zero for the residual. Also, the extra term $I$ in the gradient prevents them from becoming too small across the network, even if $f_i'(\boldsymbol{H}_{i-1})$ has elements that are approximately zero. We'll describe other potential benefits to this model towards the end of this section. 
```

chapters/cls-neural-nets.qmd
```diff
@@ -1204,25 +1239,172 @@ $$
...
 
-### Multi-Head Attention {#sec-cls-multi-attention}
+Our _query_ reflects what we are currently interested in. For example, I might make the query:
+
+> I'm interested in a `fictional` `murder-mystery` set in an `isolated location` in `French Canada` with `quirky` characters and a `duck`. 
+
+A database would `FILTER` on these values to find any rows that match the query. Hopefully, the query will return some _values_ that, in this case, correspond to names for relevant books.   
+
+The mathematics of attention mechanisms differ in their tactics, but they pursue the same strategy as the database example above.  Attention computes a quantitative similarity score for the data, usually based on dot products, which are scaled to [0, 1]. This is more of a fuzzy-matching system than an exact match would produce with a SQL `FILTER`.
```
kjell-stattenacity:
```suggestion
The mathematics of attention mechanisms differ, but all pursue the same strategy as the database example above. Attention computes quantitative similarity scores between the elements of the data, usually from dot products, and maps these scores to the interval $[0, 1]$. The result is a form of fuzzy matching rather than the exact matching produced by a SQL `FILTER`. 
```

chapters/cls-neural-nets.qmd
```diff
@@ -1204,25 +1239,172 @@ $$
...
+:::
+
+In essence, we are updating our data on a per-query basis.  A different query would generate different weights. As we'll see shortly, attention weights focus the computations on what is relevant for each individual query. 
+
+How is attention incorporated into neural networks? The attention mechanism takes incoming embeddings and produces a set of improved embeddings in a supervised manner. This means that our attention weights need to be learnable parameters so that gradient descent can optimize them. 
+
+Attention is usually followed by one or more layers of traditional MLP-type hidden units that take the preceding embedding and make it more predictive of the outcome. Many neural networks typically use _self-attention_ where $\boldsymbol{X}$ is the query, key, and value elements. 
+
+::: {.mathy-box}
+There is a lot of linear algebra on the horizon. In the next section, we'll eventually bring things back down to Earth and demonstrate attention with a small-scale example.
```
kjell-stattenacity:
```suggestion
There is a lot of linear algebra on the horizon. In the next section, we'll eventually bring things back down to earth and demonstrate attention with a small-scale example. 
```

chapters/cls-neural-nets.qmd
```diff
@@ -1204,25 +1239,172 @@ $$
...
+
+Why does self-attention use $\boldsymbol{X}$ for the query, keys, and values? The idea is that, at some point in the network's architecture, we have the current set of embeddings (e.g., the computed hidden units). These are fixed in the sense that their parameters are the same across all samples predicted by the model. With attention, each sample's representation is informed by its own internal feature context. The only type of data used here is the embeddings; the outcome or any other data are not (directly) used. Attention will refocus our embeddings based on the context of each sample^[A shorter answer to the original question is that the individual embeddings are being compared to, and updated by, themselves.]. 
+
+The attention block estimates three sets of weight matrices, all of which are $m \times m$, that project each column's embedding into a query, key, and value: 
+
+$$
+\begin{align}
+\boldsymbol{Q} &= \boldsymbol{X}\boldsymbol{W}^{q} \notag \\
```
kjell-stattenacity: I'm not familiar with the common notation for the block estimates.  However, the superscript on W makes it look like the matrix is being exponentiated rather than a notation denoting which weight matrix it belongs to.  Could we put the indicator as a subscript instead?  But please disregard this comment if the common notation is to put the indicator as a superscript.

kjell-stattenacity (commented):
A few wording suggestions and questions

chapters/cls-neural-nets.qmd
topepo: Yep!

